PR #11, opened 2024-01-24
:seedling: reorg the topics

skeeey:
## Summary

reorg the topics to:

- SourceEvents topic (`sources/+/clusters/+/sourceevents`) is a topic for sources to publish their resource create/update/delete events or status resync events
- AgentEvents topic (`sources/+/clusters/+/agentevents`) is a topic for agents to publish their resource status update events or spec resync events
- (Optional) SourceBroadcast (`sources/+/sourceevents`)  is a topic for a source to publish its events to all agents
- (Optional) AgentBroadcast (`clusters/+/agentevents`) is a topic for a agent to publish its events to all sources

### DIFF
--- a/pkg/cloudevents/generic/interface.go
+++ b/pkg/cloudevents/generic/interface.go
@@ -54,7 +54,12 @@ type Codec[T ResourceObject] interface {
 
 type CloudEventsClient[T ResourceObject] interface {
 	// Resync the resources of one source/agent by sending resync request.
-	Resync(ctx context.Context, listOptions types.ListOptions) error
+	// The second parameter is used to specify cluster name/source ID for a source/agent.
+	//   - A source sends the resource status resync request to a cluster with the given cluster name.
+	//     If setting this parameter to `types.ClusterAll`, the source will broadcast the resync request to all clusters.
+	//   - An agent sends the resources spec resync request to a source with the given source ID.
+	//     If setting this parameter to `types.SourceAll`, the agent will broadcast the resync request to all sources.
+	Resync(context.Context, string) error
 
 	// Publish the resources spec/status event to the broker.
 	Publish(ctx context.Context, eventType types.CloudEventsType, obj T) error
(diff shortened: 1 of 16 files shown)


clyang82:
Could you append the description about how many topics we have for status and the responsible for each status?

skeeey:
> Could you append the description about how many topics we have for status and the responsible for each status?

added, more details can be found the  comments on the `Topics` structrue

pkg/cloudevents/generic/interface.go
```diff
@@ -54,7 +54,7 @@ type Codec[T ResourceObject] interface {
 
 type CloudEventsClient[T ResourceObject] interface {
 	// Resync the resources of one source/agent by sending resync request.
-	Resync(ctx context.Context, listOptions types.ListOptions) error
+	Resync(context.Context, string) error
```
qiujian16, line 57: comment on what is the second variable

pkg/cloudevents/generic/types/types.go
```diff
@@ -82,16 +79,47 @@ const (
...
+	//     resync request
+	// The topic format is `^([a-z]+)/([a-z0-9-]+|\+)/([a-z]+)/\+/agentevents$`, e.g. sources/+/clusters/+/agentevents
+	// or sources/source1/clusters/+/agentevents
+	AgentEvents string `json:"agentEvents" yaml:"agentEvents"`
+
+	// SourceBroadcast is an optional topic, it is for a source to publish its events to all agents, currently, we use
+	// this topic to resync resource status from all agents for a source that does not known the exact agents, e.g.
+	//   - A source uses this topic to publish its resource status resync request with its sourceID to all the agents
+	//   - Each agent subscribes to this topic to response sources resource status resync request
+	// The topic format is `^([a-z]+)/([a-z0-9-]+|\+)/sourceevents$`, e.g. sources/source1/sourceevents or
```
qiujian16, line 113: maybe agentbroadcast/sourcebroadcast So it is different from agentevents/sourceevents

pkg/cloudevents/generic/interface.go
skeeey: added

pkg/cloudevents/generic/types/types.go
skeeey: updated

skeeey:
we may also need enable this option in the https://github.com/open-cluster-management-io/sdk-go/blob/main/pkg/cloudevents/work/clientbuilder.go#L58

PR #13, opened 2024-01-30
:bug: support to get source from share topic

skeeey:
## Related issue(s)

Fixes #

### DIFF
--- a/pkg/cloudevents/generic/options/mqtt/options.go
+++ b/pkg/cloudevents/generic/options/mqtt/options.go
@@ -250,12 +250,17 @@ func validateTopics(topics *types.Topics) error {
 }
 
 func getSourceFromEventsTopic(topic string) (string, error) {
-	subTopics := strings.Split(topic, "/")
-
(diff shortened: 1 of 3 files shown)


pkg/cloudevents/generic/options/mqtt/options.go
```diff
@@ -252,8 +252,18 @@ func validateTopics(topics *types.Topics) error {
 func getSourceFromEventsTopic(topic string) (string, error) {
 	subTopics := strings.Split(topic, "/")
 
+	// get source form share topic, e.g. $share/group/sources/+/consumers/+/agentevents
+	if subTopics[0] == "$share" {
+		if len(subTopics) != 7 {
```
qiujian16, line 257: the number is hard coded? what if the user use a different topic? can we use regex?

skeeey: The topic length is fixed, but using regex is more grace, will update this

qiujian16 (commented):
/approve
/lgtm

PR #17, opened 2024-02-28
🐛 fix resource version conversion.

morvencao:
## Summary

This PR tries to fix the `resourceversion` convert issue in cloudevent decoder, the original issue is as follow:

```bash
failed to decode cloudevent: failed to get resourceversion extension: cannot convert 1 to string
```

## Related issue(s)

Fixes #


skeeey:
LGTM

/assign @qiujian16

PR #27, opened 2024-04-09
:bug: When discarding a cloudevents client, clean up its associated resources

yanmxa:
## Summary

We need to make sure the cloudevents client releases the resources when deprecating the current client! 

Since the [client](https://github.com/cloudevents/sdk-go/blob/main/v2/client/client.go#L25-L51) hasn't provided the related behaviors. We will leverage the cloudevents [protocol](https://cloudevents.github.io/sdk-go/protocol_implementations.html#protocol-interfaces) to achieve that instead of using the client directly

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/cloudevents/generic/baseclient.go
+++ b/pkg/cloudevents/generic/baseclient.go
@@ -36,11 +36,7 @@ type baseClient struct {
 
 func (c *baseClient) connect(ctx context.Context) error {
 	var err error
-	c.cloudEventsProtocol, err = c.cloudEventsOptions.Protocol(ctx)
-	if err != nil {
-		return err
-	}
-	c.cloudEventsClient, err = cloudevents.NewClient(c.cloudEventsProtocol)
+	c.cloudEventsClient, err = c.newCloudEventsClient(ctx)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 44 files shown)


pkg/cloudevents/generic/baseclient.go
```diff
@@ -67,6 +73,13 @@ func (c *baseClient) connect(ctx context.Context) error {
 					continue
 				}
 
+				cloudEventsClient, err = cloudevents.NewClient(c.cloudEventsProtocol)
+				if err != nil {
+					runtime.HandleError(fmt.Errorf("the cloudevents client initialize failed, %v", err))
+					<-wait.RealTimer(delayFn()).C()
```
skeeey, line 79: Under what case will this error occur? if we retry, dose it can be recover?

pkg/cloudevents/generic/baseclient.go
```diff
@@ -94,6 +107,11 @@ func (c *baseClient) connect(ctx context.Context) error {
 				c.sendReceiverSignal(stopReceiverSignal)
 
 				cloudEventsClient = nil
+				err = c.cloudEventsProtocol.Close(ctx)
```
skeeey, line 110: I think you need close the protocol before setting the client to nil

pkg/cloudevents/generic/baseclient.go
yanmxa: It will occur when the client tries to connect to a crashed broker, or it hasn't been authenticated to connect the broker. 
Well, for the network changes the retry might work. But for some other issues like authentication, I think it won't work

pkg/cloudevents/generic/baseclient.go
yanmxa: Done.

qiujian16 (commented):
/approve

pkg/cloudevents/generic/baseclient.go
skeeey: nit: would you use a function to do this? thus in future, we can check the error at one place
```
fun newCloudEventsClient() client, error {
    protocol, err = get protocol
   client, err = new client with protocol
   return client err
}
```

PR #28, opened 2024-04-10
🌱 enhance resource version check for event based manifestwork patch. 

morvencao:
## Summary

remove version check for event based work client patch operation.

## Related issue(s)

ref: https://issues.redhat.com/browse/ACM-10959

### DIFF
--- a/pkg/cloudevents/generic/agentclient.go
+++ b/pkg/cloudevents/generic/agentclient.go
@@ -282,7 +282,23 @@ func (c *CloudEventAgentClient[T]) specAction(source string, obj T) (evt types.R
 		return types.Deleted, nil
 	}
 
-	if obj.GetResourceVersion() == lastObj.GetResourceVersion() {
+	// if both the current and the last object have the resource version "0", then object
+	// is considered as modified, the message broker guarantees the order of the messages
+	if obj.GetResourceVersion() == "0" && lastObj.GetResourceVersion() == "0" {
+		return types.Modified, nil
+	}
+
+	resourceVersion, err := strconv.ParseInt(obj.GetResourceVersion(), 10, 64)
+	if err != nil {
+		return evt, err
+	}
+
+	lastResourceVersion, err := strconv.ParseInt(lastObj.GetResourceVersion(), 10, 64)
+	if err != nil {
+		return evt, err
+	}
+
+	if resourceVersion <= lastResourceVersion {
 		return evt, nil
 	}
 
--- a/pkg/cloudevents/work/agent/handler/resourcehandler.go
+++ b/pkg/cloudevents/work/agent/handler/resourcehandler.go
@@ -2,11 +2,9 @@ package handler
 
 import (
 	"fmt"
-	"strconv"
 
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/watch"
-	"k8s.io/klog/v2"
 	workv1lister "open-cluster-management.io/api/client/work/listers/work/v1"
 
 	workv1 "open-cluster-management.io/api/work/v1"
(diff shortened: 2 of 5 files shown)


pkg/cloudevents/work/source/client/manifestwork.go
```diff
@@ -202,18 +194,26 @@ func (c *ManifestWorkSourceClient) Patch(ctx context.Context, name string, pt ku
...
+// If the generation is not set or invalid, then -1 is retuned. In such cases, a persistent storage
+// (e.g., a database for Maestro) is necessary to maintain the generation.
+// However, if persistent storage is unavailable, you must set the generation in the annotation
+// to ensure that the work is updated by the latest request.
+// TODO: Handle for case where the generation is not set or invalid, and no persistent storage.
+func getWorkGeneration(work *workv1.ManifestWork) int64 {
 	generation, ok := work.Annotations[common.CloudEventsGenerationAnnotationKey]
 	if !ok {
-		return -1, fmt.Errorf("the annotation %s is not found from work %s", common.CloudEventsGenerationAnnotationKey, work.UID)
+		return -1
```
skeeey, line 208: need a log here

pkg/cloudevents/work/source/client/manifestwork.go
```diff
@@ -202,18 +194,26 @@ func (c *ManifestWorkSourceClient) Patch(ctx context.Context, name string, pt ku
...
 	}
 
 	generationInt, err := strconv.Atoi(generation)
 	if err != nil {
-		return -1, err
+		return -1
```
skeeey, line 213: I think you need to keep this error, this means the generation is invalid

pkg/cloudevents/work/source/client/manifestwork.go
skeeey: or we can have an other annotation to indicate  the generation will be maintained  by remote server?

morvencao: updated this based on latest discussion.

pkg/cloudevents/work/source/client/manifestwork.go
morvencao: updated this based on latest discussion.

pkg/cloudevents/generic/agentclient.go
```diff
@@ -282,7 +282,13 @@ func (c *CloudEventAgentClient[T]) specAction(source string, obj T) (evt types.R
...
 	}
 
-	if obj.GetResourceVersion() == lastObj.GetResourceVersion() {
+	// if both the current and the last object have the resource version "0", then object
+	// is considered as modified, the message broker guarantees the order of the messages
+	if obj.GetResourceVersion() == "0" && lastObj.GetResourceVersion() == "0" {
+		return types.Modified, nil
+	}
+
+	if obj.GetResourceVersion() <= lastObj.GetResourceVersion() {
```
skeeey, line 291: GetResourceVersion returns a string? can we compare strings with this?

pkg/cloudevents/work/source/client/manifestwork.go
```diff
@@ -202,15 +197,19 @@ func (c *ManifestWorkSourceClient) Patch(ctx context.Context, name string, pt ku
...
 	generation, ok := work.Annotations[common.CloudEventsGenerationAnnotationKey]
 	if !ok {
-		return -1, fmt.Errorf("the annotation %s is not found from work %s", common.CloudEventsGenerationAnnotationKey, work.UID)
+		return 0, nil
 	}
 
 	generationInt, err := strconv.Atoi(generation)
 	if err != nil {
-		return -1, err
+		return -1, fmt.Errorf("failed to convert generation %s to int: %v", generation, err)
```
qiujian16, line 212: 0?

pkg/cloudevents/work/agent/handler/resourcehandler.go
```diff
@@ -39,9 +39,13 @@ func NewManifestWorkAgentHandler(lister workv1lister.ManifestWorkNamespaceLister
 				return fmt.Errorf("failed to parse the resourceVersion of the manifestwork %s, %v", lastWork.Name, err)
 			}
 
-			if resourceVersion <= lastResourceVersion {
-				klog.Infof("The work %s resource version is less than or equal to cached, ignore", work.Name)
-				return nil
+			// if both the current and the last object have the resource version "0", then object
+			// is considered as changed, the message broker guarantees the order of the messages
+			if resourceVersion != 0 || lastResourceVersion != 0 {
```
qiujian16, line 44: I do not think the check is necessary.

pkg/cloudevents/work/source/client/manifestwork.go
morvencao: updated to `0`. actually, the error takes precedence, so it doesn't matter.

pkg/cloudevents/work/agent/handler/resourcehandler.go
skeeey: agree, we already check the resourceVersion in the agent client, this should be redundant, we can remove the all resourceVersion check here

morvencao: removed this version check in handler.

pkg/cloudevents/generic/agentclient.go
skeeey: I think we need to convert string to int firstly, then compare them

morvencao: done.

qiujian16 (commented):
/approve
/lgtm

PR #34, opened 2024-04-26
:sparkles: support client cert rotating (mqtt)

skeeey:
## Related issue(s)

Fixes #


pkg/cloudevents/generic/options/mqtt/options.go
```diff
@@ -28,19 +29,53 @@ const (
...
+type MQTTDialer struct {
+	TLSConfig  *tls.Config
+	BrokerHost string
+	Timeout    time.Duration
+
+	conn net.Conn
+}
+
+func (d *MQTTDialer) Dial() (net.Conn, error) {
+	if d.TLSConfig != nil {
```
qiujian16, line 41: can we use the same mechanism as here https://github.com/kubernetes/client-go/blob/master/transport/transport.go#L378?

skeeey: yes, we used here, see https://github.com/open-cluster-management-io/sdk-go/pull/34/files#diff-ae2df3cd963e3abffbbda90110244c84dc73f795649a7c200c1979dd2a311276R185

qiujian16: then why we still need a go routine to periodically check?

skeeey: I think the whole cert rotation includes two parts:
1. a go routine to check the cert changes, if the cert was changed, it closes the old connection to trigger the client reconnect with new certs (see https://github.com/kubernetes/client-go/blob/master/transport/cache.go#L118, https://github.com/kubernetes/client-go/blob/master/transport/cert_rotation.go#L53)
2. a way to load the cert dynamically ( https://github.com/kubernetes/client-go/blob/master/transport/transport.go#L378)

qiujian16: oh so the established connection cannot directly pick the cert in tlsConfig?

skeeey: yes, it is

skeeey: the periodical go routine is needed, my test has some problem, it is bothering me ...

PR #36, opened 2024-05-06
✨ [WIP]add garbage collector for cloudevents work client.

morvencao:
## Summary

This PR adds Garbage Collector for cloudevents work client.

## Related issue(s)

Fixes #


go.mod
```diff
@@ -70,6 +70,7 @@ require (
 	gopkg.in/inf.v0 v0.9.1 // indirect
 	gopkg.in/tomb.v1 v1.0.0-20141024135613-dd632973f1e7 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
+	k8s.io/apiextensions-apiserver v0.29.0 // indirect
```
qiujian16, line 73: why it depends on apiextension-apiserver?

test/integration/crds/0000_00_addon.open-cluster-management.io_clustermanagementaddons.crd.yaml
```diff
@@ -0,0 +1,677 @@
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
```
qiujian16, line 3: from the test perspective, adding these crds is not necessary, you could use built-in resource for testing.

pkg/cloudevents/work/garbagecollector/garbagecollector.go
```diff
@@ -0,0 +1,538 @@
...
+	attemptToDelete workqueue.RateLimitingInterface
+}
+
+// NewGarbageCollector creates a new garbage collector instance.
+func NewGarbageCollector(
+	workClient workv1.WorkV1Interface,
+	metadataClient metadata.Interface,
+	workInformer workv1informers.ManifestWorkInformer,
+	ownerGVRFilters map[schema.GroupVersionResource]*metav1.ListOptions) *GarbageCollector {
+	return &GarbageCollector{
```
qiujian16, line 109: I unerstand this is from GC in k8s, but I actually think this is a bit too complicated. My understanding is you just need to build an index for each manifestwork based on its owner. @skeeey do we support indexer with cloudevent based informer?

And then watch the other resources, caculate the index and find related manifestwork.

go.mod
morvencao: checked this again.
we don't need this directly.
it is imported by the [envtest](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/envtest) package used for testing only.

test/integration/crds/0000_00_addon.open-cluster-management.io_clustermanagementaddons.crd.yaml
morvencao: removed this `crds` firectory, I just realized we can use CRDs from vendor directory, just like what we do in [ocm](https://github.com/open-cluster-management-io/ocm/blob/main/test/integration/work/suite_test.go#L55) repo.

go.mod
qiujian16: I do not think you need envtest if we do not add addon crd.

test/integration/crds/0000_00_addon.open-cluster-management.io_clustermanagementaddons.crd.yaml
qiujian16: may point is we do not need to test based on addon. You could have other resources like configmap/secret for testing.

pkg/cloudevents/work/garbagecollector/garbagecollector.go
skeeey: I think the indexer should be supported, for cloudevent based informer, we did not do any  additional operation when we build the informer, just like k8s does

morvencao: @qiujian16 Added indexer for manifestwork by owner resources to simplify the code, see: https://github.com/open-cluster-management-io/sdk-go/pull/36/commits/a14632b2a76b5e379fb14212655bf60f7eff4ddd
Please take another look?

pkg/cloudevents/work/garbagecollector/garbagecollector.go
```diff
@@ -0,0 +1,324 @@
...
+
+// HasSynced returns true if any monitors exist AND all those monitors'
+// controllers HasSynced functions return true.
+func (gc *GarbageCollector) hasSynced(logger klog.Logger) bool {
+	if len(gc.monitors) == 0 {
+		logger.V(4).Info("garbage collector monitors are not synced: no monitors")
+		return false
+	}
+
+	for resource, monitor := range gc.monitors {
```
qiujian16, line 212: also need to check check if work informer has synced.

pkg/cloudevents/work/garbagecollector/garbagecollector.go
```diff
@@ -0,0 +1,324 @@
...
+	logger.Info("Attempting to delete manifestwork", "ownerUID", dep.ownerUID, "namespacedName", dep.namespacedName)
+
+	latest, err := gc.getManifestwork(ctx, dep.namespacedName)
+	if err != nil {
+		return requeueItem
+	}
+
+	// may happen when manifestwork deletion failed in last attempt
+	ownerReferences := latest.GetOwnerReferences()
+	if len(ownerReferences) == 0 {
```
qiujian16, line 270: not quite sure how this can happen

pkg/cloudevents/work/garbagecollector/garbagecollector.go
```diff
@@ -0,0 +1,324 @@
...
+
+func (gc *GarbageCollector) attemptToDeleteWorker(ctx context.Context, item interface{}) workQueueItemAction {
+	dep, ok := item.(*dependent)
+	if !ok {
+		utilruntime.HandleError(fmt.Errorf("expect a *dependent, got %v", item))
+		return forgetItem
+	}
+
+	logger := klog.FromContext(ctx)
+	logger.Info("Attempting to delete manifestwork", "ownerUID", dep.ownerUID, "namespacedName", dep.namespacedName)
```
qiujian16, line 261: needs a hight log level

pkg/cloudevents/work/garbagecollector/operations.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"context"
+	"encoding/json"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	workapiv1 "open-cluster-management.io/api/work/v1"
+)
+
+func (gc *GarbageCollector) getManifestwork(ctx context.Context, namespacedName types.NamespacedName) (*workapiv1.ManifestWork, error) {
```
qiujian16, line 12: why we need this wrapper? seems not necessary

pkg/cloudevents/work/garbagecollector/gc.go
```diff
@@ -0,0 +1,208 @@
+package garbagecollector
+
+// import (
```
qiujian16, line 3: can we just remove this file?

pkg/cloudevents/work/garbagecollector/garbagecollector.go
```diff
@@ -0,0 +1,324 @@
...
+			return requeueItem
+		}
+		if _, err = gc.patchManifestwork(ctx, dep.namespacedName, jmp, types.MergePatchType); err != nil {
+			logger.Error(err, "Failed to patch manifestwork with json patch")
+			return requeueItem
+		}
+		logger.V(4).Info("Successfully removed owner reference from manifestwork", "owner", dep.ownerUID, "manifestwork", dep.namespacedName)
+
+		// if the deleted owner reference is the only owner reference then delete the manifestwork
+		if len(ownerReferences) == 1 {
```
qiujian16, line 301: hrm, I thin we can delete at first then remove ownerrefs  right?

pkg/cloudevents/work/garbagecollector/garbagecollector_test.go
```diff
@@ -0,0 +1,287 @@
...
+	"k8s.io/client-go/rest"
+
+	addonapiv1alpha1 "open-cluster-management.io/api/addon/v1alpha1"
+	workclientset "open-cluster-management.io/api/client/work/clientset/versioned"
+	fakework "open-cluster-management.io/api/client/work/clientset/versioned/fake"
+	workinformers "open-cluster-management.io/api/client/work/informers/externalversions"
+	workapiv1 "open-cluster-management.io/api/work/v1"
+)
+
+func buildWork(namespace, name string, owners []string) *workapiv1.ManifestWork {
```
qiujian16, line 31: I think we have a similar func in common testing utitlity already?

go.mod
morvencao: removed envtest dependency.

test/integration/crds/0000_00_addon.open-cluster-management.io_clustermanagementaddons.crd.yaml
morvencao: updated integration test to use configmap/secret as owner resources.

pkg/cloudevents/work/garbagecollector/garbagecollector.go
morvencao: the workInformer is also added to monitors in `startMonitors()`, so the it will be also checked here.

pkg/cloudevents/work/garbagecollector/garbagecollector.go
morvencao: added log level for this.

pkg/cloudevents/work/garbagecollector/operations.go
morvencao: removed the wrapper and this file.

pkg/cloudevents/work/garbagecollector/gc.go
morvencao: removed this file.

pkg/cloudevents/work/garbagecollector/garbagecollector_test.go
morvencao: it looks like we don't have func to set owner refs for the manifestwork.

pkg/cloudevents/work/garbagecollector/garbagecollector.go
morvencao: yes, we can.
updated the order to delete the manifestwork firstly when it's the only owner.

pkg/cloudevents/work/garbagecollector/garbagecollector.go
morvencao: This is related to https://github.com/open-cluster-management-io/sdk-go/pull/36#discussion_r1606179748
Sence we change the delete manifestwork order, I think we don't need this any more, so I removed this check.

pkg/cloudevents/work/garbagecollector/garbagecollector.go
```diff
@@ -0,0 +1,342 @@
...
+	ownerGVRFilters map[schema.GroupVersionResource]*metav1.ListOptions
+	// each monitor list/watches a resource (including manifestwork)
+	monitors monitors
+	// garbage collector attempts to delete the items in attemptToDelete queue when the time is ripe.
+	attemptToDelete workqueue.RateLimitingInterface
+}
+
+// NewGarbageCollector creates a new garbage collector instance.
+func NewGarbageCollector(
+	workClient workv1.WorkV1Interface,
```
qiujian16, line 71: since this will only be used in cloudevent case, can we use the clientHolder as the input which already include client and informer I think. And do we need metaClient as the input? What the expected interface that we want developer to use?

qiujian16 (commented):
the code looks good in general, only need to consider the interface provided to the developer.

pkg/cloudevents/work/garbagecollector/garbagecollector.go
morvencao: updated to use clientHolder to initialize GC.
metadataclient is used to create informer for owner resources.
Now we have three parameters to create GC:

```bash
func NewGarbageCollector(
	workClientHolder *cloudeventswork.ClientHolder,
	metadataClient metadata.Interface,
	ownerGVRFilters map[schema.GroupVersionResource]*metav1.ListOptions,
) *GarbageCollector
```

pkg/cloudevents/work/garbagecollector/garbagecollector.go
```diff
@@ -0,0 +1,344 @@
...
+	}
+
+	logger.V(4).Info("Started monitors", "started", started, "total", len(gc.monitors))
+	return nil
+}
+
+// monitorFor creates monitor for owner resource
+func (gc *GarbageCollector) monitorFor(logger klog.Logger, gvr schema.GroupVersionResource, listOptions *metav1.ListOptions) (*monitor, error) {
+	handlers := cache.ResourceEventHandlerFuncs{
+		AddFunc:    func(obj interface{}) {},
```
skeeey, line 159: We may miss one corner case, when a controller is restarting, during this time, the work owner is deleted, when the work status send back, currently, we just requeue this work, so this work will not be deleted

so do we need to add this work back if the work has owner and check the work owner here?

qiujian16: I think we have not consider the work that only have status in this PR either, right? Might need add a TODO

skeeey: yes, I think so

morvencao: thanks for the suggestion, added a TODO for this.

pkg/cloudevents/work/garbagecollector/garbagecollector.go
```diff
@@ -0,0 +1,346 @@
...
+
+			ownerUID := accessor.GetUID()
+			objs, err := gc.workIndexer.ByIndex(manifestWorkByOwner, string(ownerUID))
+			if err != nil {
+				utilruntime.HandleError(fmt.Errorf("failed to get manifestwork by owner UID index: %v", err))
+				return
+			}
+
+			for _, o := range objs {
+				manifestWork := o.(*workapiv1.ManifestWork)
```
skeeey, line 182: nit: checking the converting result

pkg/cloudevents/work/garbagecollector/garbagecollector.go
```diff
@@ -0,0 +1,346 @@
...
+		utilruntime.HandleError(fmt.Errorf("expect a *dependent, got %v", item))
+		return forgetItem
+	}
+
+	logger := klog.FromContext(ctx)
+	logger.V(4).Info("Attempting to delete manifestwork", "ownerUID", dep.ownerUID, "namespacedName", dep.namespacedName)
+
+	latest, err := gc.workClient.ManifestWorks(dep.namespacedName.Namespace).Get(ctx, dep.namespacedName.Name, metav1.GetOptions{})
+
+	if err != nil {
```
skeeey, line 270: if one work is not found here, can we forget it?

pkg/cloudevents/work/garbagecollector/garbagecollector.go
morvencao: added check, although the convert should always succeed.

pkg/cloudevents/work/garbagecollector/garbagecollector.go
morvencao: reasonable to forget it if this happens.

PR #40, opened 2024-05-09
:sparkles: Introduce options builder

clyang82:
## Summary

Introduce an option builder to Encapsulate the concrete implementation. so the user only needs to provide the type and configuration path, the build returns the`CloudEventsSourceOptions`. it can get rid of the potential dependencies for kafka.

## Related issue(s)

Fixes #


pkg/constants/constants.go
```diff
@@ -0,0 +1,8 @@
+package constants
```
qiujian16, line 1: pkg/cloudevents/constants

pkg/cloudevents/work/configloader.go
```diff
@@ -49,7 +44,7 @@ func (l *ConfigLoader) WithKubeConfig(kubeConfig *rest.Config) *ConfigLoader {
 // TODO using a specified config instead of any
 func (l *ConfigLoader) LoadConfig() (string, any, error) {
 	switch l.configType {
-	case ConfigTypeKube:
+	case constants.ConfigTypeKube:
```
qiujian16, line 47: if we have the builder, why do we still need the switch here?

pkg/cloudevents/generic/optionsbuilder.go
```diff
@@ -0,0 +1,39 @@
...
+
+	"open-cluster-management.io/sdk-go/pkg/cloudevents/generic/options"
+	"open-cluster-management.io/sdk-go/pkg/cloudevents/generic/options/grpc"
+	"open-cluster-management.io/sdk-go/pkg/cloudevents/generic/options/kafka"
+	"open-cluster-management.io/sdk-go/pkg/cloudevents/generic/options/mqtt"
+	"open-cluster-management.io/sdk-go/pkg/constants"
+)
+
+// BuildCloudEventsSourceOptions builds the cloudevents source options based on the broker type
+func BuildCloudEventsSourceOptions(messageBrokerType, messageBrokerConfigPath, clientId, sourceId string) (*options.CloudEventsSourceOptions, error) {
```
skeeey, line 14: I think we may need two build functions
1.  different source config object -> CloudEventsSourceOptions
2. different agent config object -> CloudEventsAgentOptions

we may refactor the https://github.com/open-cluster-management-io/sdk-go/blob/main/pkg/cloudevents/work/clientbuilder.go#L124 and https://github.com/open-cluster-management-io/sdk-go/blob/main/pkg/cloudevents/work/clientbuilder.go#L140 and put them here

pkg/cloudevents/work/configloader.go
skeeey: this function help us to load the config object from config file, and then the work builder build the CloudEventsSourceOptions/CloudEventsAgentOptions with the config object

pkg/cloudevents/generic/optionsbuilder.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+// BuildCloudEventsSourceOptions builds the cloudevents source options based on the broker type
+func BuildCloudEventsSourceOptions(config any, clientId, sourceId string) (*options.CloudEventsSourceOptions, error) {
+	switch config := config.(type) {
+	case *mqtt.MQTTOptions:
+		return mqtt.NewSourceOptions(config, clientId, sourceId), nil
+	case *grpc.GRPCOptions:
+		return grpc.NewSourceOptions(config, sourceId), nil
+	case *map[string]interface{}:
```
skeeey, line 19: nit: would we also define an KafkaOptions instead of `map[string]interface{}`

pkg/constants/constants.go
clyang82: Done

pkg/cloudevents/work/configloader.go
clyang82: Done

pkg/cloudevents/generic/optionsbuilder.go
clyang82: Done

skeeey:
LGTM

pkg/cloudevents/work/configloader.go
qiujian16: can't we put this in the newly builder file, instead of the kube type. So that we do not need to add switch in multiple files.

clyang82: Do you mean combine `loadConfig` with `CloudeventsOptions`?

qiujian16: or at least they are in the same places

clyang82: Done

PR #41, opened 2024-05-11
:seedling: Transfer cloudevent message with structured-mode

yanmxa:
## Summary

use [structured-mode](https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#message) to transfer messages, so that the customer can receive the whole event(data and metadata) from the protocol payload.

## Related issue(s)

Reference: https://github.com/stolostron/multicluster-global-hub/issues/919

### DIFF
--- a/pkg/cloudevents/generic/baseclient.go
+++ b/pkg/cloudevents/generic/baseclient.go
@@ -124,6 +124,7 @@ func (c *baseClient) publish(ctx context.Context, evt cloudevents.Event) error {
 			latency, evt))
 	}
 
+	ctx = cloudevents.WithEncodingStructured(ctx)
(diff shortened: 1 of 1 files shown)


pkg/cloudevents/generic/baseclient.go
```diff
@@ -124,6 +124,7 @@ func (c *baseClient) publish(ctx context.Context, evt cloudevents.Event) error {
 			latency, evt))
 	}
 
+	ctx = cloudevents.WithEncodingStructured(ctx)
```
skeeey, line 127: what does this mean?

yanmxa: It used to wrap all the cloudevent in the payload, maybe the maestro doesn't need it. I will close it and we can reopen it if it's necessary in the future

PR #37, opened 2024-05-06
:seedling: Add the kafka  integration test

yanmxa:
## Summary

Add Kafka integration test for Cloudevent protocol

## Related issue(s)

Fixes # https://issues.redhat.com/browse/ACM-11002

PR #42, opened 2024-05-11
:seedling: Add the kafka integration test

yanmxa:
## Summary

Add Kafka integration test for Cloudevent protocol

Related issue(s)

Fixes # https://issues.redhat.com/browse/ACM-11002, https://github.com/open-cluster-management-io/sdk-go/pull/37

### DIFF
--- a/pkg/cloudevents/generic/options/kafka/agentoptions.go
+++ b/pkg/cloudevents/generic/options/kafka/agentoptions.go
@@ -31,10 +31,7 @@ func NewAgentOptions(kafkaOptions *KafkaOptions, clusterName, agentID string) *o
 		errorChan:    make(chan error),
 	}
 
-	groupID, err := kafkaOptions.ConfigMap.Get("group.id", "")
-	if groupID == "" || err != nil {
-		_ = kafkaOptions.ConfigMap.SetKey("group.id", agentID)
-	}
+	_ = kafkaOptions.ConfigMap.SetKey("group.id", agentID)
 
 	return &options.CloudEventsAgentOptions{
 		CloudEventsOptions: kafkaAgentOptions,
--- a/pkg/cloudevents/generic/options/kafka/agentoptions.go
+++ b/pkg/cloudevents/generic/options/kafka/agentoptions.go
@@ -31,7 +31,10 @@ func NewAgentOptions(kafkaOptions *KafkaOptions, clusterName, agentID string) *o
 		errorChan:    make(chan error),
 	}
 
-	_ = kafkaOptions.ConfigMap.SetKey("group.id", agentID)
+	groupID, err := kafkaOptions.ConfigMap.Get("group.id", "")
+	if groupID == "" || err != nil {
+		_ = kafkaOptions.ConfigMap.SetKey("group.id", agentID)
+	}
 
 	return &options.CloudEventsAgentOptions{
 		CloudEventsOptions: kafkaAgentOptions,
--- a/test/integration/cloudevents/cloudevents_kafka_test.go
+++ b/test/integration/cloudevents/cloudevents_kafka_test.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"time"
 
+	kafkav2 "github.com/confluentinc/confluent-kafka-go/v2/kafka"
 	jsonpatch "github.com/evanphx/json-patch"
 	"github.com/onsi/ginkgo"
 	"github.com/onsi/gomega"
@@ -17,6 +18,7 @@ import (
 
 	workv1 "open-cluster-management.io/api/work/v1"
 	"open-cluster-management.io/sdk-go/pkg/cloudevents/generic"
+	"open-cluster-management.io/sdk-go/pkg/cloudevents/generic/options/kafka"
 	kafkaoptions "open-cluster-management.io/sdk-go/pkg/cloudevents/generic/options/kafka"
 	"open-cluster-management.io/sdk-go/pkg/cloudevents/generic/types"
 	"open-cluster-management.io/sdk-go/pkg/cloudevents/work"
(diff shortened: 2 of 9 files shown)


test/integration/cloudevents/cloudevents_kafka_test.go
```diff
@@ -0,0 +1,172 @@
...
+			if err != nil {
+				return err
+			}
+			if !meta.IsStatusConditionTrue(storeResource.Status.Conditions, "Created") {
+				return fmt.Errorf("unexpected status %v", storeResource.Status.Conditions)
+			}
+			return nil
+		}, 10*time.Second, 1*time.Second).Should(gomega.Succeed())
+
+		ginkgo.By("Agent resync resource from source")
```
clyang82, line 136: Your case is for resync. it is for sync-up a new resource. Do you plan to add a test case for re-sync? if so, you need to restart or create a new work client. I am not sure it is necessary.

yanmxa: Add a agent restart case now

clyang82 (commented):
LGTM

test/integration/cloudevents/suite_test.go
```diff
@@ -19,19 +20,25 @@ import (
...
-var mqttOptions *mqtt.MQTTOptions
-var mqttSourceCloudEventsClient generic.CloudEventsClient[*source.Resource]
-var grpcServer *source.GRPCServer
-var grpcOptions *grpcoptions.GRPCOptions
-var grpcSourceCloudEventsClient generic.CloudEventsClient[*source.Resource]
-var eventBroadcaster *source.EventBroadcaster
-var store *source.MemoryStore
-var consumerStore *source.MemoryStore
+var (
+	mqttBroker                  *mochimqtt.Server
```
qiujian16, line 30: we have to refactor this long list in the future. We might build a brokerInterface with different implementations. which could help construct the test better. Let's add a todo here.

yanmxa: Done.

test/integration/cloudevents/cloudevents_kafka_test.go
```diff
@@ -0,0 +1,219 @@
...
+			NewAgentClientHolder(newAgentCtx)
+		gomega.Expect(err).ToNot(gomega.HaveOccurred())
+
+		// Note: Different configuration for the new agent will have different behavior
+		//   Case1: new agentID(group.id) + "auto.offset.reset": latest
+		//        The agent has to wait until the consumer is ready to send message, like time.Sleep(5 * time.Second)
+		//   Case2: keep the same agentID for the new agent, it will ignore the "auto.offset.reset" automatically
+		//        Then we don't need wait the consumer ready, cause it will consume message from last committed
+		//        But the agent will wait a long time(test result is 56 seconds) to receive the message
+		time.Sleep(5 * time.Second)
```
qiujian16, line 182: I am not so sure what this sleep mean. Does it imply that we cannot start informer immediately? If so will it actually be a bug?

pkg/cloudevents/generic/options/kafka/agentoptions.go
```diff
@@ -31,10 +31,7 @@ func NewAgentOptions(kafkaOptions *KafkaOptions, clusterName, agentID string) *o
 		errorChan:    make(chan error),
 	}
 
-	groupID, err := kafkaOptions.ConfigMap.Get("group.id", "")
-	if groupID == "" || err != nil {
-		_ = kafkaOptions.ConfigMap.SetKey("group.id", agentID)
-	}
+	_ = kafkaOptions.ConfigMap.SetKey("group.id", agentID)
```
skeeey, line 34: does this mean we will not support set the group.id outside, it will be always agentID?

test/integration/cloudevents/cloudevents_kafka_test.go
yanmxa: The sleep is to make sure that the agent's consumer is ready, and then the agent can send a resync event to the source server.

Otherwise, the agent may lose events resynced from the source if its consumer is starting...

pkg/cloudevents/generic/options/kafka/agentoptions.go
yanmxa: It means we will always define the `group.id` explicitly. It will be `agentID` for the agent and `sourceID` for the source server.

yanmxa: The reason to keep the `group.id` will always be `sourceID`/`agentID` can guarantee the continuity of their(source/agent) consumption. Let save the agent is crashed, and if we keep the same `agentID` to restart the agent, it will consume messages from its last received.

test/integration/cloudevents/cloudevents_kafka_test.go
yanmxa: We can also remove the `time.Sleep`. And it comes to the [case 2](https://github.com/open-cluster-management-io/sdk-go/pull/42/files/3b7c3990264bf2e37a55bf3fbad33baffd2f424c#diff-f20a8669d58cc9b0405730a3d6ab499d73dde40c1dbe1080d6aee65d58f2c11cR176-R181). Then we have to wait a longer time(50 seconds after testing) to receive the resynced events from source

qiujian16: is this to fasten the testing only? we need to clarify that. It is weird we have to sleep to fasten the testing.

yanmxa: Yes! In a real environment. We don't need to wait for the consumer to be ready. It's only for the mocked Kafka cluster in the test.

pkg/cloudevents/generic/options/kafka/agentoptions.go
skeeey: we already provide the `groud.id` config in https://github.com/open-cluster-management-io/sdk-go/blob/main/pkg/cloudevents/generic/options/kafka/options.go#L45

if we use this change, do we still need the `groud.id` in the config? or we don't set the `group.id` here, we require the `group.id` in the config, or we don't change the original code

yanmxa: Let's make it simple! And get rid of the `groud.id` from the config. What do you think of it?

skeeey: I think the original code may be more suitable :)

yanmxa: Done!

PR #50, opened 2024-05-27
:sparkles: add cache for source manifestwork client

skeeey:
## Related issue(s)

https://github.com/open-cluster-management-io/ocm/issues/476


pkg/cloudevents/work/clientbuilder.go
```diff
@@ -113,126 +109,121 @@ func (b *ClientHolderBuilder) WithInformerConfig(
...
+		if len(b.sourceID) == 0 {
+			return nil, nil, nil, fmt.Errorf("source id is required")
+		}
+
+		options, err := generic.BuildCloudEventsSourceOptions(config, b.clientID, b.sourceID)
+		if err != nil {
+			return nil, nil, nil, err
+		}
+
+		localCache := newCacheFunc(ctx)
```
qiujian16, line 199: why we need to newCacheFun in here instead of outside this func?

pkg/cloudevents/work/clientbuilder.go
```diff
@@ -113,126 +109,121 @@ func (b *ClientHolderBuilder) WithInformerConfig(
...
 
-// NewAgentClientHolder returns a ClientHolder for agent
-func (b *ClientHolderBuilder) NewAgentClientHolder(ctx context.Context) (*ClientHolder, error) {
-	switch config := b.config.(type) {
-	case *rest.Config:
-		return b.newKubeClients(config)
-	default:
-		options, err := generic.BuildCloudEventsAgentOptions(config, b.clusterName, b.clientID)
-		if err != nil {
+	if err := b.initWorkClientWatcherCache(ctx, localCache); err != nil {
```
qiujian16, line 129: we do not need to init when client is kube type client, right?

pkg/cloudevents/work/clientbuilder.go
```diff
@@ -113,126 +109,121 @@ func (b *ClientHolderBuilder) WithInformerConfig(
...
+		localCache := newCacheFunc(ctx)
+		clientSet, cloudEventsClient, err := newCloudEventsSourceClient(ctx, b.sourceID, options, localCache, b.codecs)
+
+		return clientSet, cloudEventsClient, localCache, err
 	}
+}
 
-	workLister := &ManifestWorkLister{}
-	watcher := watcher.NewManifestWorkWatcher()
+func newCloudEventsSourceClient(
```
qiujian16, line 206: what is the purpose of this func and what is the different from newSourceClient? we should add some comment. It is not quite clear to me

pkg/cloudevents/work/clientinformerbuilder.go
```diff
@@ -0,0 +1,135 @@
...
+	agentclient "open-cluster-management.io/sdk-go/pkg/cloudevents/work/agent/client"
+	agenthandler "open-cluster-management.io/sdk-go/pkg/cloudevents/work/agent/handler"
+	"open-cluster-management.io/sdk-go/pkg/cloudevents/work/cache"
+	"open-cluster-management.io/sdk-go/pkg/cloudevents/work/internal"
+	"open-cluster-management.io/sdk-go/pkg/cloudevents/work/watcher"
+)
+
+// NewSourceClientHolderWithInformer returns a ClientHolder with a ManifestWorkInformer for a source.
+//
+// This function should be used in scenarios where a ManifestWorkInformer is required
```
qiujian16, line 27: Does it mean the informer has to be started after this func, otherwise the client doest not work?

test/integration/cloudevents/manifestworksourceclient_cache_test.go
```diff
@@ -0,0 +1,388 @@
+package cloudevents
+
+import (
+	"context"
```
qiujian16, line 4: I think we should have a test on how this works with customized lister.

pkg/cloudevents/work/clientbuilder.go
skeeey: this cache only is needed by CloudEvents based client, but the NewSourceClient is able to create kube client, so I call this function only when the cache is need

pkg/cloudevents/work/clientbuilder.go
skeeey: right, we do not need to init the kube type client

pkg/cloudevents/work/clientbuilder.go
skeeey: `newSourceClient` can return both kube client and cloudevents based client (by invoking the `newSourceClient `), and the `newCloudEventsSourceClient` only return the cloudevents based client, I will add some comments

pkg/cloudevents/work/clientinformerbuilder.go
skeeey: yes, informer should be started, I will add comments for this

test/integration/cloudevents/manifestworksourceclient_cache_test.go
machi1990: +1

machi1990:
Thanks @skeeey & @qiujian16 for working on this.

I reviewed the PR, it touches some points we discussed in our today call and I've summarised the challenges in this discussion issue https://github.com/open-cluster-management-io/ocm/issues/483

pkg/cloudevents/work/clientbuilder.go
skeeey: merge the `newCloudEventsSourceClient` to `newSourceClient` to eliminate the confusion

test/integration/cloudevents/manifestworksourceclient_cache_test.go
skeeey: added a case for customized lister https://github.com/open-cluster-management-io/sdk-go/pull/50/files#diff-824eab8018ea74b205a8ca714b906070347b418d5ade2913511793f388bb93d1R386

pkg/cloudevents/work/clientbuilder.go
skeeey: put this after `cloudEventsClient != nil` to make this more clear

pkg/cloudevents/work/clientbuilder.go
```diff
@@ -113,158 +108,130 @@ func (b *ClientHolderBuilder) WithInformerConfig(
...
+//
+// Note: the ListLocalWorksFunc should only returns the works that have status. After the cloudevents based source
+// manifestwork client is created, the source client will resync the status of these works.
+func (b *ClientHolderBuilder) WithListLocalWorksFunc(lister ListLocalWorksFunc) *ClientHolderBuilder {
+	b.localWorkLister = lister
+	return b
+}
+
+// NewSourceClientHolder returns a ClientHolder for a source
 func (b *ClientHolderBuilder) NewSourceClientHolder(ctx context.Context) (*ClientHolder, error) {
```
clyang82, line 122: Would you like to add a test for this NewSourceClientHolder with a faked localworkList?

skeeey: The test already added with https://github.com/open-cluster-management-io/sdk-go/pull/50/files#diff-824eab8018ea74b205a8ca714b906070347b418d5ade2913511793f388bb93d1R386

skeeey: oh, the file is folded, it is at test/integration/cloudevents/manifestworksourceclient_cache_test.go `Using the source client with customized lister`

pkg/cloudevents/work/clientbuilder.go
```diff
@@ -113,158 +108,130 @@ func (b *ClientHolderBuilder) WithInformerConfig(
...
-	switch config := b.config.(type) {
-	case *rest.Config:
-		return b.newKubeClients(config)
-	default:
-		options, err := generic.BuildCloudEventsSourceOptions(config, b.clientID, b.sourceID)
-		if err != nil {
+	workClientSet, cloudEventsClient, cloudEventsClientWatcherCache, err := b.newSourceClient(ctx, cache.NewLocalCache)
```
qiujian16, line 123: why not switch on rest.Config outside newSourceCient func?

pkg/cloudevents/work/clientinformerbuilder.go
```diff
@@ -0,0 +1,137 @@
...
+)
+
+// NewSourceClientHolderWithInformer returns a ClientHolder with a ManifestWorkInformer for a source.
+// This function is suitable for the scenarios where a ManifestWorkInformer is required
+//
+// Note: the ClientHolder should be used after the ManifestWorkInformer starts
+func (b *ClientHolderBuilder) NewSourceClientHolderWithInformer(
+	ctx context.Context,
+) (*ClientHolder, workv1informers.ManifestWorkInformer, error) {
+	workClientSet, cloudEventsClient, watcherCache, err := b.newSourceClient(ctx, cache.NewInformerWatcherCache)
```
qiujian16, line 32: I still think we should initCache outside of newSourceClient

pkg/cloudevents/work/clientbuilder.go
skeeey: Do you means we only focus on the cloudevetns based client here? for kube type client, user will not use this to build, right?

clyang82:
You need to update this document - https://github.com/open-cluster-management-io/sdk-go/blob/main/pkg/cloudevents/README.md

pkg/cloudevents/work/clientbuilder.go
skeeey: now, I only build cloudevetns based client here

skeeey:
> You need to update this document - https://github.com/open-cluster-management-io/sdk-go/blob/main/pkg/cloudevents/README.md

yeah, I will udpate the readme after this pr is done

pkg/cloudevents/work/clientinformerbuilder.go
skeeey: I think the problem here, we need wait for the cache is ready, then do resync, if we allow initCache outside of newSourceClient, could we also allow to resync outside of newSourceClient? 

This means, we only resync when client reconnected, the sdk-go will not resync at the  client start, the user will do resync or not by themselves

skeeey: I tied to initialize the cache with cache self, now I use withclientstore instead of withworklist, PTAL

pkg/cloudevents/work/store/informer.go
```diff
@@ -0,0 +1,58 @@
...
+func (c *InformerWatcherStore) Delete(work *workv1.ManifestWork) error {
+	c.LocalWatcherStore.result <- watch.Event{Type: watch.Deleted, Object: work}
+	return nil
+}
+
+func (c *InformerWatcherStore) HasInitiated() bool {
+	return c.LocalWatcherStore.initiated
+}
+
+func (c *InformerWatcherStore) SetStore(store cache.Store) {
```
qiujian16, line 55: I think we can set store in NewInformerWatcherStore? This is required for this store, isn't it?

pkg/cloudevents/work/store/local.go
```diff
@@ -0,0 +1,306 @@
...
+	done   chan struct{}
+
+	store      cache.Store
+	eventQueue cache.Queue
+
+	initiated bool
+}
+
+// NewLocalWatcherStore returns a LocalWatcherStore
+func NewLocalWatcherStore(ctx context.Context) *LocalWatcherStore {
```
qiujian16, line 48: is this necessary, why a user wants to use this instead of NewLocalWatcherStoreWithWorks?

pkg/cloudevents/work/store/store.go
```diff
@@ -0,0 +1,56 @@
...
+// This is useful for determining if caches have synced.
+type StoreInitiated func() bool
+
+// WorkClientWatcherStore extends the watch interface with a work store.
+type WorkClientWatcherStore interface {
+	watch.Interface
+
+	Add(work *workv1.ManifestWork) error
+
+	Update(work *workv1.ManifestWork) error
```
qiujian16, line 23: why store needs update/delete? This should be readonly.

pkg/cloudevents/work/clientbuilder.go
```diff
@@ -113,129 +112,68 @@ func (b *ClientHolderBuilder) WithInformerConfig(
...
-		options, err := generic.BuildCloudEventsAgentOptions(config, b.clusterName, b.clientID)
-		if err != nil {
-			return nil, err
-		}
-		return b.newAgentClients(ctx, options)
-	}
+// WithResync control the client resync (Default is true), if it's true, the resync happens when
+//  1. after the client's store is initiated
+//  2. the client reconnected
+func (b *ClientHolderBuilder) WithResync(resync bool) *ClientHolderBuilder {
```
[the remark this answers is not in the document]
skeeey: updated

pkg/cloudevents/work/clientbuilder.go
```diff
@@ -113,129 +112,68 @@ func (b *ClientHolderBuilder) WithInformerConfig(
...
-	manifestWorkClient.SetLister(manifestWorkLister)
 
-	sourceHandler := sourcehandler.NewManifestWorkSourceHandler(manifestWorkLister, watcher)
-	cloudEventsClient.Subscribe(ctx, sourceHandler.HandlerFunc())
+	if !b.resync {
+		return &ClientHolder{workClientSet: workClientSet}, nil
+	}
 
-	go sourceHandler.Run(ctx.Done())
+	// start a go route to resync the works when this client reconnected
```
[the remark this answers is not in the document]
skeeey: corrected

pkg/cloudevents/work/store/informer.go
skeeey: This because we need get a store from an informer, but for creating an informer we need provide a client, so we have to put the informer store later, like this https://github.com/open-cluster-management-io/sdk-go/pull/50/files#diff-5eed0f48418509bcf24930171650c37d57f8e7d309415fc44ced658098731322R194

pkg/cloudevents/work/store/local.go
skeeey: I think this for the case that the user don't have a local env (e.g. db) to save the works, they will create a work client without init works, does this make sense?

pkg/cloudevents/work/store/store.go
skeeey: I think these are not readonly, the client need to maintain its works in this store

pkg/cloudevents/work/store/local.go
skeeey: And I used this store as the default store for the client

I think it may be better to ask user to provide a store when creating a client than giving a default store?

qiujian16: That will be confusing I think. A client without initList func would be buggy.

pkg/cloudevents/work/store/store.go
qiujian16: if we have a implementation of maestro, how they wonld implement Add/Update here?

skeeey: they need send the watched event to the watcher result like the informer watcher store

pkg/cloudevents/work/store/local.go
skeeey: OK, only keep `func NewLocalWatcherStore(ctx context.Context, listFunc ListLocalWorksFunc)` and ask to provide a store

pkg/cloudevents/work/store/store.go
skeeey: there is an example for this https://github.com/skeeey/practices/blob/main/go/cmd/example/maestro/store.go

qiujian16: we should add some comment in the add/update/delete func, it does not really need to update a store, but just send an watch event.

skeeey: added

PR #52, opened 2024-06-03
🐛 fix agent resyncs the deleted spec from source

morvencao:
## Related issue(s)

https://github.com/open-cluster-management-io/ocm/issues/492

### DIFF
--- a/pkg/cloudevents/generic/sourceclient.go
+++ b/pkg/cloudevents/generic/sourceclient.go
@@ -229,6 +229,14 @@ func (c *CloudEventSourceClient[T]) respondResyncSpecRequest(
 	}
 
 	for _, obj := range objs {
+		// respond with the deleting resource regardless of the resource version
+		if !obj.GetDeletionTimestamp().IsZero() {
+			if err := c.Publish(ctx, eventType, obj); err != nil {
+				return err
+			}
+			continue
+		}
+
 		lastResourceVersion := findResourceVersion(string(obj.GetUID()), resourceVersions.Versions)
 		currentResourceVersion, err := strconv.ParseInt(obj.GetResourceVersion(), 10, 64)
 		if err != nil {
@@ -239,13 +247,6 @@ func (c *CloudEventSourceClient[T]) respondResyncSpecRequest(
 		// the version of the work is not maintained on source or the source's work is newer than agent, send
 		// the newer work to agent
 		if currentResourceVersion == 0 || currentResourceVersion > lastResourceVersion {
-			if err := c.Publish(ctx, eventType, obj); err != nil {
-				return err
-			}
-		}
-
-		// respond with the deleting resource regardless of the resource version
-		if !obj.GetDeletionTimestamp().IsZero() {
 			if err := c.Publish(ctx, eventType, obj); err != nil {
 				return err
 			}
--- a/pkg/cloudevents/generic/sourceclient.go
+++ b/pkg/cloudevents/generic/sourceclient.go
@@ -243,6 +243,13 @@ func (c *CloudEventSourceClient[T]) respondResyncSpecRequest(
 				return err
 			}
 		}
+
+		// respond with the deleting resource regardless of the resource version
+		if !obj.GetDeletionTimestamp().IsZero() {
+			if err := c.Publish(ctx, eventType, obj); err != nil {
+				return err
+			}
+		}
 	}
 
 	// the resources do not exist on the source, but exist on the agent, delete them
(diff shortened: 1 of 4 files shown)


pkg/cloudevents/work/agent/codec/manifest.go
```diff
@@ -65,10 +65,6 @@ func (c *ManifestCodec) Encode(source string, eventType types.CloudEventsType, w
 		return nil, fmt.Errorf("failed to find originalsource from the work %s", work.UID)
 	}
 
-	if len(work.Spec.Workload.Manifests) != 1 {
-		return nil, fmt.Errorf("too many manifests in the work %s", work.UID)
```
qiujian16, line 69: why do we need to remove this check?

morvencao: This issue occurs when a resource is deleted on the server side during agent downtime. After the agent restarts, the source only sends CloudEvents with a deletion timestamp and an empty payload as a resync response, preventing the resource spec from being rebuilt in the agent cache.

As an alternative, we can also to make the source send the complete spec even when a resource is deleted on the source side.

/cc @skeeey

qiujian16: so it should be when deletiongTimeStamp is not zero or len(work.Spec.Workload.Manifests) != 1?

pkg/cloudevents/generic/sourceclient.go
```diff
@@ -250,6 +250,13 @@ func (c *CloudEventSourceClient[T]) respondResyncSpecRequest(
 				return err
 			}
 		}
+
+		// respond with the deleting resource regardless of the resource version
+		if !obj.GetDeletionTimestamp().IsZero() {
```
skeeey, line 255: add this before the resource version check

I think this because when we deleting a resource on the source, the resource version will not be added?

pkg/cloudevents/generic/sourceclient.go
```diff
@@ -250,6 +250,13 @@ func (c *CloudEventSourceClient[T]) respondResyncSpecRequest(
...
+			if err := c.Publish(ctx, eventType, obj); err != nil {
+				return err
+			}
```
skeeey, line 258: also need `continue`

pkg/cloudevents/work/agent/codec/manifest.go
```diff
@@ -65,8 +65,11 @@ func (c *ManifestCodec) Encode(source string, eventType types.CloudEventsType, w
...
+	// for status update cases other than manifest deletion, there should be only one manifest in the work
```
skeeey, line 68: nit: may need add reason why we don't check the manifests of a deleted work

pkg/cloudevents/generic/sourceclient.go
morvencao: move this before version check.
right, for deletion case, version will not be changed.

pkg/cloudevents/generic/sourceclient.go
morvencao: added `continue`.

pkg/cloudevents/work/agent/codec/manifest.go
morvencao: done.

pkg/cloudevents/work/agent/codec/manifest.go
qiujian16: yes, this is fine

morvencao: added some comments in https://github.com/open-cluster-management-io/sdk-go/pull/52/commits/23816953d64e053f653c432a41e6d0ac9a998f2a

pkg/cloudevents/work/agent/codec/manifest.go
```diff
@@ -146,6 +146,7 @@ func (c *ManifestCodec) Decode(evt *cloudevents.Event) (*workv1.ManifestWork, er
 			return nil, fmt.Errorf("failed to get deletiontimestamp, %v", err)
 		}
 
+		work.Finalizers = []string{workv1.ManifestWorkFinalizer}
```
skeeey, line 149: we need some comments here

morvencao: added comments  in https://github.com/open-cluster-management-io/sdk-go/pull/52/commits/02ec4756d6ae1f61111d41598dbdfae13cd5a03f

morvencao:
rebased the code to fix the verify testing.

PR #53, opened 2024-06-03
:sparkles: add servingcert controller

zhiweiyin318:
## Related issue(s)

Fixes #


.gitignore
```diff
@@ -20,3 +20,6 @@
 # Go workspace file
 go.work
 .idea
+deps.diff
```
qiujian16, line 23: why these files.

pkg/servingcert/controller.go
```diff
@@ -0,0 +1,207 @@
...
+
+	c.namespacedInformers.Start(ctx.Done())
+	go controller.Run(ctx, 1)
+}
+
+func (c *ServingCertController) sync(ctx context.Context, syncCtx factory.SyncContext, key string) error {
+	return c.SyncOne(ctx)
+}
+
+func (c *ServingCertController) SyncOne(ctx context.Context) error {
```
qiujian16, line 179: why do we need a SyncOne here? Are you try to call it only once?

pkg/servingcert/controller.go
```diff
@@ -0,0 +1,207 @@
...
+	targetCertValidity, resyncInterval time.Duration
+	kubeClient                         kubernetes.Interface
+	namespacedInformers                informers.SharedInformerFactory
+	signingRotation                    certrotation.SigningRotation
+	caBundleRotation                   certrotation.CABundleRotation
+	targetRotations                    []certrotation.TargetRotation
+}
+
+func NewServingCertController(namespace string, kubeClient kubernetes.Interface) *ServingCertController {
+	namespacedInformers := informers.NewSharedInformerFactoryWithOptions(kubeClient, 10*time.Minute, informers.WithNamespace(namespace))
```
qiujian16, line 46: if we put informer here, we cannot filter by name

pkg/certrotation/target.go
```diff
@@ -53,10 +58,11 @@ func (c TargetRotation) EnsureTargetCertKeyPair(signingCertKeyPair *crypto.CA, c
 		return err
 	}
 
-	if _, _, err = helpers.ApplySecret(context.TODO(), c.Client, targetCertKeyPairSecret); err != nil {
+	if targetCertKeyPairSecret, _, err = helpers.ApplySecret(context.TODO(), c.Client, targetCertKeyPairSecret); err != nil {
 		return err
 	}
-	return nil
+
+	return c.loadTargetSecret(targetCertKeyPairSecret)
```
qiujian16, line 65: I would check 
```
if c.LoadDir == "" {
		return nil
	}
```
here

.gitignore
zhiweiyin318: deps.diff is generated after run `make verify`, tls.key and tls.crt are left over if integration tests fail.

pkg/servingcert/controller.go
zhiweiyin318: in the controller-runtime operator with webhook case,  the controller-runtime manger needs a loaded serving cert secret to start, so need to SyncOne before start.

pkg/servingcert/controller.go
zhiweiyin318: the informer only watches the resources in the specified ns, maybe we do not need to filter by name?

pkg/certrotation/target.go
zhiweiyin318: done

pkg/servingcert/controller.go
qiujian16: what if we set strict rbac.

pkg/servingcert/controller.go
qiujian16: I think we need to document this usage.

pkg/servingcert/controller.go
zhiweiyin318: changed to that one informer only watches one resource.

PR #56, opened 2024-06-07
:seedling: set work meta data to ManifestBundle event on source

skeeey:
## Related issue(s)

Fixes #

### DIFF
--- a/pkg/cloudevents/work/source/codec/manifestbundle.go
+++ b/pkg/cloudevents/work/source/codec/manifestbundle.go
@@ -1,6 +1,7 @@
 package codec
 
 import (
+	"encoding/json"
 	"fmt"
 	"strconv"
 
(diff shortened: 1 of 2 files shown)


pkg/cloudevents/work/source/codec/manifestbundle.go
```diff
@@ -15,6 +16,20 @@ import (
...
+	ExtensionWorkName = "name"
+
+	// ExtensionWorkLabels is an extension attribute for work labels.
+	ExtensionWorkLabels = "labels"
+
+	// ExtensionWorkAnnotations is an extension attribute for work annotations.
+	ExtensionWorkAnnotations = "annotations"
+
+	// ExtensionWorkFinalizers is an extension attribute for work finalizers.
+	ExtensionWorkFinalizers = "finalizers"
```
qiujian16, line 30: so what about we set whole metaObject as one value?

skeeey: let me think about this...

skeeey: using the whole metaObject, PTAL

qiujian16 (commented):
/approve
/lgtm

PR #59, opened 2024-06-17
:bug: always receive the reconnected signal to avoid client blocking

skeeey:
## Related issue(s)

Fixes #


qiujian16:
so if we disable resync, the reconnect will block without this fix, right?

skeeey:
> so if we disable resync, the reconnect will block without this fix, right?

yes, after the client reconnected, the client will send a reconnect signal to a channel, it will be blocked if there is no receiver

PR #60, opened 2024-06-17
:seedling: add querying options to list and watch

skeeey:
## Related issue(s)

Fixes #


pkg/cloudevents/work/agent/client/manifestwork.go
```diff
@@ -70,7 +70,7 @@ func (c *ManifestWorkAgentClient) Get(ctx context.Context, name string, opts met
 
 func (c *ManifestWorkAgentClient) List(ctx context.Context, opts metav1.ListOptions) (*workv1.ManifestWorkList, error) {
 	klog.V(4).Infof("list manifestworks")
-	works, err := c.watcherStore.List(opts)
+	works, err := c.watcherStore.List(c.clusterName, opts)
```
qiujian16, line 73: does it support all?

skeeey: yes, it does

skeeey: handle here https://github.com/open-cluster-management-io/sdk-go/pull/60/files#diff-74d6100c69d0c83ceae8a3b90070967958787cbc0106de6ec2cdd0157828bb40R97

skeeey: but for agent part, the cluster name always should have a given value

pkg/cloudevents/work/agent/client/manifestwork.go
```diff
@@ -26,6 +26,7 @@ import (
 type ManifestWorkAgentClient struct {
 	cloudEventsClient *generic.CloudEventAgentClient[*workv1.ManifestWork]
 	watcherStore      store.WorkClientWatcherStore
+	namespace         string
```
qiujian16, line 29: I think namespace is the same as clusterName? why do we need two var?

skeeey: The clusterName is initialized when we create this client, when we build this client, we will start to subscribe to a given cluster

namespace is set when we use this client, e.g. `client.namespace("aaaa").list(...)`

skeeey: it should be ok to remove `clusterName`,  if we give a wrong namespace for client, the related work will not be found

skeeey: updated to only use namespace

qiujian16: add a comment here since it should be the same as clusterName.

skeeey: added

qiujian16 (commented):
/approve
/lgtm

PR #62, opened 2024-06-21
:bug: handle grpc idle status

skeeey:
## Summary

If the grpc client is connected, when the grpc connections is down, the grpc connection state will change from Ready to Idle, the `TRANSIENT_FAILURE` seems only happened when establishing the connection


qiujian16:
needs some comments

PR #63, opened 2024-06-27
:seedling: provide a normalize library to generate scores for AddOnPlacementScore

haoqing0110:
## Summary
Ref: https://github.com/open-cluster-management-io/ocm/issues/68

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/apis/cluster/v1alpha1/score.go
+++ b/pkg/apis/cluster/v1alpha1/score.go
@@ -1,0 +1,43 @@
+package v1alpha1
+
+import (
+	"k8s.io/klog/v2"
+)
+
+// MaxScore is the upper bound of the normalized score.
+const MaxScore = 100
+
+// MinScore is the lower bound of the normalized score.
+const MinScore = -100
+
+// scoreNormalizer holds the minimum and maximum values for normalization,
+// provides a normalize library to generate scores for AddOnPlacementScore.
+type scoreNormalizer struct {
+	min float64
+	max float64
+}
+
+// NewScoreNormalizer creates a new instance of scoreNormalizer with given min and max values.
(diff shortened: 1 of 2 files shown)


pkg/apis/cluster/v1alpha1/score.go
```diff
@@ -0,0 +1,43 @@
...
+		min: min,
+		max: max,
+	}
+}
+
+// Normalize normalizes a given value to the range -100 to 100 based on the min and max values.
+func (s *scoreNormalizer) Normalize(value float64) (score int64, err error) {
+	if value > s.max {
+		// If the value exceeds the maximum, set score to MaxScore.
+		score = int64(MaxScore)
```
qiujian16, line 32: why we need to define MinScore as float and convert to int here?

haoqing0110: the MinScore and MaxScore should also used to normalize the score when s.min < value < s.max, just update the formula on line 38.

pkg/apis/cluster/v1alpha1/score.go
```diff
@@ -0,0 +1,43 @@
...
+		score = MaxScore
+		// If the value is less than or equal to the minimum, set score to MinScore.
+	} else if value <= s.min {
+		score = MinScore
+	} else {
+		// Otherwise, normalize the value to the range -100 to 100.
+		score = (int32)((MaxScore-MinScore)*(value-s.min)/(s.max-s.min) + MinScore)
+	}
+
+	klog.Infof("value = %v, min = %v, max = %v, score = %v", value, s.min, s.max, score)
```
qiujian16, line 41: V(2)

haoqing0110: fixed.

PR #69, opened 2024-07-16
:seedling: using kube api errors for work clients

skeeey:
## Related issue(s)

Fixes #

### DIFF
--- a/pkg/cloudevents/work/source/client/manifestwork.go
+++ b/pkg/cloudevents/work/source/client/manifestwork.go
@@ -3,6 +3,7 @@ package client
 import (
 	"context"
 	"fmt"
+	"net/http"
 
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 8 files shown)


qiujian16 (commented):
/approve

/assign @morvencao

pkg/cloudevents/work/source/client/manifestwork.go
```diff
@@ -170,24 +185,32 @@ func (c *ManifestWorkSourceClient) List(ctx context.Context, opts metav1.ListOpt
...
+		return nil, errors.NewInternalError(err)
+	}
+	if !exists {
+		return nil, errors.NewNotFound(common.ManifestWorkGR, name)
 	}
 
 	patchedWork, err := utils.Patch(pt, lastWork, data)
 	if err != nil {
-		return nil, err
+		return nil, errors.NewInternalError(err)
```
morvencao, line 213: Nit: Could we make the patch error messages more fine-grained? For example, specify if the error is due to field conflicts or other issues.

skeeey: for conflicts, it seems hard to judge on the client part, but we can change this error with BadRequest, I think this more reasonable

morvencao (commented):
/lgtm

pkg/cloudevents/work/source/client/manifestwork.go
morvencao: ok, it makes sense.

PR #70, opened 2024-07-19
✨ enable grpc keepalive.

morvencao:
## Summary

This PR introduces keepalive parameters to the cloudevents client using gRPC. By default, these parameters are disabled. Users can optionally enable it to detect broken gRPC connections. MQTT has similar feature to detect the half-open connection.

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/cloudevents/generic/options/grpc/options.go
+++ b/pkg/cloudevents/generic/options/grpc/options.go
@@ -14,6 +14,7 @@ import (
 	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/credentials/insecure"
 	"google.golang.org/grpc/credentials/oauth"
+	"google.golang.org/grpc/keepalive"
 	"gopkg.in/yaml.v2"
 
 	"open-cluster-management.io/sdk-go/pkg/cloudevents/generic/options"
@@ -22,11 +23,20 @@ import (
 
 // GRPCOptions holds the options that are used to build gRPC client.
 type GRPCOptions struct {
-	URL            string
-	CAFile         string
-	ClientCertFile string
-	ClientKeyFile  string
-	TokenFile      string
+	URL              string
+	CAFile           string
+	ClientCertFile   string
+	ClientKeyFile    string
+	TokenFile        string
+	KeepAliveOptions KeepAliveOptions
+}
+
+// KeepAliveOptions holds the keepalive options for the gRPC client.
+type KeepAliveOptions struct {
+	Enable              bool
+	Time                time.Duration
+	Timeout             time.Duration
+	PermitWithoutStream bool
 }
 
 // GRPCConfig holds the information needed to build connect to gRPC server as a given user.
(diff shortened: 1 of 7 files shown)


pkg/cloudevents/generic/options/grpc/options.go
```diff
@@ -36,6 +46,26 @@ type GRPCConfig struct {
...
+	// A minimum value of 10s is enforced if set below that. Default is 30s.
+	Time *time.Duration `json:"time,omitempty" yaml:"time,omitempty"`
+
+	// Timeout sets the duration the client waits for a response after a keepalive ping.
+	// If no response is received, the connection is closed. Default is 10s.
+	Timeout *time.Duration `json:"timeout,omitempty" yaml:"timeout,omitempty"`
+
+	// PermitWithoutStream determines if keepalive pings are sent when there are no active RPCs.
+	// If false, pings are not sent and Time and Timeout are ignored. Default is false.
+	PermitWithoutStream bool `json:"permitWithoutStream,omitempty" yaml:"permitWithoutStream,omitempty"`
```
skeeey, line 68: what is the different with the enable? if we set enable=true and permitwithoutstream=false, what happens?

pkg/cloudevents/generic/options/grpc/options.go
```diff
@@ -36,6 +46,26 @@ type GRPCConfig struct {
...
 	ClientKeyFile string `json:"clientKeyFile,omitempty" yaml:"clientKeyFile,omitempty"`
+	// keepalive options
+	KeepAliveConfig KeepAliveConfig `json:"keepAliveConfig,omitempty" yaml:"keepAliveConfig,omitempty"`
+}
+
+// KeepAliveConfig holds the keepalive options for the gRPC client.
+type KeepAliveConfig struct {
+	// Enable specifies whether the keepalive option is enabled.
+	// When disabled, other keepalive configurations are ignored. Default is false.
+	Enable bool `json:"enable,omitempty" yaml:"enable,omitempty"`
```
skeeey, line 57: Under what cases should we enable this?

pkg/cloudevents/generic/options/grpc/options.go
morvencao: enable is global flag for keep alive.
if we set enable=true and permitwithoutstream=false, the keep alive still is opened, but it will not send ping if no rpc streaming.

pkg/cloudevents/generic/options/grpc/options.go
morvencao: If we want to quick detect the broken connection (for example, the broken connection caused by broken network), we can enable this.

When we enable the keepalive, client will send keeepalive pings, if we can't receive ping response for a certain duration, the client will close the connection (state from `READY` to `IDLE`).

skeeey: so for our case, we may enable this by default?

pkg/cloudevents/generic/options/grpc/options.go
skeeey: so we may enable this by default? because we generally need a subscription

morvencao: Will raise another PR in maestro to enable this.

pkg/cloudevents/generic/options/grpc/options.go
morvencao: Will raise another PR in maestro to enable this.

pkg/cloudevents/generic/options/grpc/options.go
```diff
@@ -22,11 +23,20 @@ import (
...
+	URL              string
+	CAFile           string
+	ClientCertFile   string
+	ClientKeyFile    string
+	TokenFile        string
+	KeepAliveOptions KeepAliveOptions
+}
+
+// KeepAliveOptions holds the keepalive options for the gRPC client.
+type KeepAliveOptions struct {
```
qiujian16, line 35: I am a bit lost on why we need both option and config.

morvencao: Config is used to build options, similar to kubeconfig, in YAML/JSON format, and can be mounted via a configmap or secret. Options can also be created without config, such as for testing.

PR #71, opened 2024-08-06
:seedling: change watchstore list return value type to supporting list continue

skeeey:
## Related issue(s)

Fixes #


qiujian16 (commented):
/approve
/lgtm

PR #72, opened 2024-08-06
✨ support token based authn for cloudevents grpc option.

morvencao:
## Summary

This PR tries to add token based authentication for cloudevents grpc option.

## Related issue(s)

Fixes #


test/integration/cloudevents/util/cert.go
```diff
@@ -155,3 +156,43 @@ func SignClientCert(caCert *x509.Certificate, caKey *rsa.PrivateKey, d time.Dura
...
+	}))
+}
+
+func WriteTokenToTempFile(token string) (string, error) {
+	// create temp file and write token to it
+	return WriteToTempFile([]byte(token))
+}
+
+// WriteToTempFile writes data to a temporary file and returns the file path.
+func WriteToTempFile(data []byte) (string, error) {
```
qiujian16, line 174: These are quite common util func, would you check if there exist one already or we could move these func to a common util package.

pkg/cloudevents/generic/options/grpc/options.go
```diff
@@ -61,12 +66,16 @@ func BuildGRPCOptionsFromFlags(configPath string) (*GRPCOptions, error) {
 	if config.ClientCertFile != "" && config.ClientKeyFile != "" && config.CAFile == "" {
 		return nil, fmt.Errorf("setting clientCertFile and clientKeyFile requires caFile")
 	}
+	if config.TokenFile != "" && config.CAFile == "" {
+		return nil, fmt.Errorf("setting tokenFile requires caFile")
```
skeeey, line 70: so if we use the token auth, the grpc server tls must be enabled?

morvencao: yes, this is a requirement for gRPC, token auth must have TLS enabled.

test/integration/cloudevents/util/cert.go
morvencao: Yes, there are many places using the same function, mostly from testing code. I’ve created a common util package and updated these findings to use this new util package.

qiujian16:
lease @skeeey for final tag

pkg/util/util.go
```diff
@@ -0,0 +1,30 @@
+package util
```
skeeey, line 1: these functions are mainly used for test, could we put these to pkg/testing dir?

morvencao: done.

PR #74, opened 2024-08-28
🐛 fix race condition between work status update and deletion.

morvencao:
## Summary

1. This PR mainly addresses a race condition where a work status update can overwrite the DeletionTimestamp, preventing the resource from being deleted. The issue occurs when an update event is followed immediately by a delete event from the source, causing the agent to skip deletion. 

    To resolve this, we now verify the resource version before updating the agent store for work status update.

3. Additionally, any update event that follows a delete event for the same resource from the source will be ignored from agent side.

4. In this PR we've also removed a duplicate debug log that printed the received event. (that is printed in [baseclient.go](https://github.com/open-cluster-management-io/sdk-go/blob/1afb1eab940dc51bf39fe81f38fac4de91a3c5ca/pkg/cloudevents/generic/baseclient.go#L161))

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/cloudevents/work/agent/client/manifestwork.go
+++ b/pkg/cloudevents/work/agent/client/manifestwork.go
@@ -137,6 +137,8 @@ func (c *ManifestWorkAgentClient) Patch(ctx context.Context, name string, pt kub
 
 	if statusUpdated {
 		eventType.Action = common.UpdateRequestAction
+		// publish the status update event to source, source will check the resource version
+		// and reject the update if it's status update is outdated.
 		if err := c.cloudEventsClient.Publish(ctx, eventType, newWork); err != nil {
 			return nil, workerrors.NewPublishError(common.ManifestWorkGR, name, err)
 		}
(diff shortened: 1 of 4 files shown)


pkg/cloudevents/work/agent/client/manifestwork.go
```diff
@@ -140,9 +140,31 @@ func (c *ManifestWorkAgentClient) Patch(ctx context.Context, name string, pt kub
...
+		lastResourceVersion, err := strconv.ParseInt(latestWork.GetResourceVersion(), 10, 64)
+		if err != nil {
+			return nil, errors.NewInternalError(err)
+		}
+		newResourceVersion, err := strconv.ParseInt(newWork.GetResourceVersion(), 10, 64)
+		if err != nil {
+			return nil, errors.NewInternalError(err)
+		}
+		// ensure the resource version of the work is not outdated
+		if newResourceVersion < lastResourceVersion {
```
qiujian16, line 163: will this be a problem we return error here? since we publish but do not update local store?

morvencao: If we only consider the deletion case, there might not be an issue because the resource will be deleted later.
Or we can move the version check before publishing the event.

skeeey: this may not be a problem, if the resource version does not match, that means
1. the resource spec has beed updated on the source
2.the resource spec has beed updated on the agent local cache

because we publish the status update with resource version, so even if the update request is sent, the source should not accept this request.

for agent, this update is not for the current newest resource, so we should also  reject it

qiujian16: we need some comment here.

morvencao:
@qiujian16 Another look?

qiujian16 (commented):
/approve
/lgtm

PR #76, opened 2024-09-04
🌱 add latency metrics for resource resync.

morvencao:
## Summary

This PR adds metrics to measure the duration of resource spec resync and status resync operations.

## Related issue(s)

Fixes #


pkg/cloudevents/generic/metrics_collector.go
```diff
@@ -0,0 +1,113 @@
...
+
+import (
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+func init() {
+	// Register the metrics for resource resync:
+	RegisterResourceResyncMetrics()
```
skeeey, line 11: we may only expose the `resourceSpecResyncDurationMetric` and `resourceStatusResyncDurationMetric`, and let sdk user to select register them or not?

And expecting the resync metrics, do we need to add other metrics in sdk level? e.g. the pub/sub?

morvencao: 1. removed `init` function to let sdk user to register metrics.
2. for other metrics, we can start by covering some in the Maestro gRPC server implementation. Additional metrics can be added later based on user feedback. comments @clyang82 ?

clyang82: agree. let us wait for feedback before emitting more metrics.

pkg/cloudevents/generic/metrics_collector.go
```diff
@@ -0,0 +1,108 @@
...
+)
+
+// metricsLabels - Array of labels added to metrics:
+var metricsLabels = []string{
+	metricsSourceLabel,
+	metricsClusterLabel,
+	metrucsDataTypeLabel,
+}
+
+// Names of the metrics:
```
qiujian16, line 26: we need more docs here on what the format of metrics and what is the expected output as an example

pkg/cloudevents/generic/metrics_collector.go
```diff
@@ -0,0 +1,108 @@
...
+const (
+	specResyncDurationMetric   = "spec_resync_duration_seconds"
+	statusResyncDurationMetric = "status_resync_duration_seconds"
+)
+
+// Description of the resource spec resync duration metric:
+var resourceSpecResyncDurationMetric = prometheus.NewHistogramVec(
```
qiujian16, line 33: can we have some tests?

pkg/cloudevents/generic/metrics_collector.go
morvencao: added some comments and sample metrics expected collected.

pkg/cloudevents/generic/metrics_collector.go
morvencao: added a simple test to check the metrics label and sum/count are as expected.

qiujian16 (commented):
/approve
/lgtm

PR #80, opened 2024-09-14
:seedling: compare status update event sequence ID in source client

skeeey:
## Summary
Compare status update event sequence ID in work source client to ensure that the messages received by the source client are in the same order as the messages sent by the agent client

This will help to guard against if some message queues cannot guarantee the order of the messages they send

### DIFF
--- a/pkg/cloudevents/work/store/base.go
+++ b/pkg/cloudevents/work/store/base.go
@@ -200,6 +200,23 @@ func (b *workProcessor) handleWork(work *workv1.ManifestWork) error {
 		return nil
 	}
 
+	if updatedWork.Annotations == nil {
+		updatedWork.Annotations = map[string]string{}
+	}
+	lastSequenceID := lastWork.Annotations[common.CloudEventsSequenceIDAnnotationKey]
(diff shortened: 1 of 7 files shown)


pkg/cloudevents/work/store/base.go
```diff
@@ -200,13 +200,31 @@ func (b *workProcessor) handleWork(work *workv1.ManifestWork) error {
...
 	}
 
+	if updatedWork.Annotations == nil {
+		updatedWork.Annotations = map[string]string{}
+	}
+	lastSequenceID := lastWork.Annotations[common.CloudEventsSequenceIDAnnotationKey]
+	sequenceID := work.Annotations[common.CloudEventsSequenceIDAnnotationKey]
+	greater, err := utils.CompareSnowflakeSequenceIDs(lastSequenceID, sequenceID)
+	if err != nil {
+		klog.Errorf("invalid resource version for work %s/%s, %v", lastWork.Namespace, lastWork.Name, err)
```
qiujian16, line 210: s/invalid resource version/invalid sequenceID

skeeey: done

PR #83, opened 2024-10-08
:seedling: avoid to print whole event in the error log

skeeey:
## Summary
If the event is too large, it will cut off the error message in the log


qiujian16:
if it is only event id, how is it valuable? Can we use it to find the specific event?

skeeey:
> if it is only event id, how is it valuable? Can we use it to find the specific event?

Using event context instead of event id, the context will include the all event meta data, e.g. id, source, type, extensions etc.

PR #84, opened 2024-10-09
:seedling: upgrade mqtt lib

skeeey:
## Summary

Before the eclipse/paho.golang v0.20, its pinger will never be a timeout, and pingFailHandler will never be called (https://github.com/eclipse/paho.golang/issues/137), it may lead to our mqtt client will not reconnect after the connection was broken


qiujian16 (commented):
/approve

would you add some comment in the PR on the issue this tries to resolve.

PR #85, opened 2024-10-10
🌱 add cloudevents metrics.

morvencao:
## Summary

This pull request adds metrics for CloudEvents to the SDK. The aim is to provide better observability for events processed by the SDK.

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/cloudevents/generic/baseclient.go
+++ b/pkg/cloudevents/generic/baseclient.go
@@ -15,7 +15,6 @@ import (
 	"k8s.io/utils/clock"
 
 	"open-cluster-management.io/sdk-go/pkg/cloudevents/generic/options"
-	"open-cluster-management.io/sdk-go/pkg/cloudevents/generic/types"
 )
 
 const (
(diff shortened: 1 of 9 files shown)


pkg/cloudevents/generic/baseclient.go
```diff
@@ -130,6 +133,14 @@ func (c *baseClient) publish(ctx context.Context, evt cloudevents.Event) error {
 		return fmt.Errorf("failed to send event %s, %v", evt.Context, result)
 	}
 
+	clusterName := evt.Context.GetExtensions()[types.ExtensionClusterName].(string)
```
skeeey, line 136: this extension may be not required for an agent?

morvencao: For agent events, the source can represent the cluster name, typically in the format `cluster1-work-agent`. Alternatively, we can leave the cluster name empty, as the source alone is sufficient to infer the event's originating cluster. Which one do you prefer?

skeeey: how about we put this part in the source/agent client? so we won't need any judgment and we also can get enough information for this action

morvencao: sounds good, I updated the code as suggested.

pkg/cloudevents/generic/baseclient.go
```diff
@@ -155,6 +158,12 @@ func (c *baseClient) subscribe(ctx context.Context, receive receiveFn) {
 				go func() {
 					if err := c.cloudEventsClient.StartReceiver(receiverCtx, func(evt cloudevents.Event) {
 						klog.V(4).Infof("Received event: %s", evt)
+						clusterName := evt.Context.GetExtensions()[types.ExtensionClusterName].(string)
+						eventType, err := types.ParseCloudEventsType(evt.Type())
+						if err == nil {
+							// only increase the received counter for the known event types
+							increaseCloudEventsReceivedCounter(evt.Source(), clusterName, eventType.CloudEventsDataType.String())
```
skeeey, line 165: we may also handle this part as the send event

morvencao: done.

PR #87, opened 2024-10-28
:bug: using local flag instead of setting client to nil

skeeey:
## Summary

Currently, we set the client to nil when a client start reconnecting, this may cause nil point

```
2024-10-17T18:53:15.218Z	ERROR	generic/baseclient.go:88	the cloudevents client is disconnected, ping handler error: PINGRESP timed out
open-cluster-management.io/sdk-go/pkg/cloudevents/generic.(*baseClient).connect.func1
	/go/pkg/mod/open-cluster-management.io/sdk-go@v0.14.1-0.20241009075853-0c5ba39fd113/pkg/cloudevents/generic/baseclient.go:88
2024-10-17T18:53:15.219Z	ERROR	generic/baseclient.go:164	failed to receive cloudevents, error while opening the inbound connection: use of closed network connection
open-cluster-management.io/sdk-go/pkg/cloudevents/generic.(*baseClient).subscribe.func1.1
	/go/pkg/mod/open-cluster-management.io/sdk-go@v0.14.1-0.20241009075853-0c5ba39fd113/pkg/cloudevents/generic/baseclient.go:164
2024-10-17T18:54:18.841Z	ERROR	generic/baseclient.go:88	the cloudevents client is disconnected, read tcp 10.132.64.71:39424->20.150.164.45:8883: use of closed network connection
open-cluster-management.io/sdk-go/pkg/cloudevents/generic.(*baseClient).connect.func1
	/go/pkg/mod/open-cluster-management.io/sdk-go@v0.14.1-0.20241009075853-0c5ba39fd113/pkg/cloudevents/generic/baseclient.go:88
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0x2056149]
```

In this pr, we use local flag to mark the client not ready

### DIFF
--- a/pkg/cloudevents/generic/baseclient.go
+++ b/pkg/cloudevents/generic/baseclient.go
@@ -41,6 +41,7 @@ type baseClient struct {
 	cloudEventsRateLimiter flowcontrol.RateLimiter
 	receiverChan           chan int
 	reconnectedChan        chan struct{}
+	clientReady            bool
 }
 
 func (c *baseClient) connect(ctx context.Context) error {
@@ -53,10 +54,10 @@ func (c *baseClient) connect(ctx context.Context) error {
 	// start a go routine to handle cloudevents client connection errors
 	go func() {
 		for {
-			if c.currentClient() == nil {
+			if !c.isClientReady() {
 				klog.V(4).Infof("reconnecting the cloudevents client")
 
-				cloudEventsClient, err := c.newCloudEventsClient(ctx)
+				c.cloudEventsClient, err = c.newCloudEventsClient(ctx)
 				// TODO enhance the cloudevents SKD to avoid wrapping the error type to distinguish the net connection
 				// errors
 				if err != nil {
@@ -65,10 +66,9 @@ func (c *baseClient) connect(ctx context.Context) error {
 					<-wait.RealTimer(DelayFn()).C()
 					continue
 				}
-				// the cloudevents network connection is back, refresh the current cloudevents client and send the
-				// receiver restart signal
+				// the cloudevents network connection is back, mark the client ready and send the receiver restart signal
 				klog.V(4).Infof("the cloudevents client is reconnected")
-				c.resetClient(cloudEventsClient)
+				c.setClientReady(true)
 				c.sendReceiverSignal(restartReceiverSignal)
 				c.sendReconnectedSignal()
 			}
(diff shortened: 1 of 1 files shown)


pkg/cloudevents/generic/baseclient.go
```diff
@@ -52,11 +53,13 @@ func (c *baseClient) connect(ctx context.Context) error {
 
 	// start a go routine to handle cloudevents client connection errors
 	go func() {
+		retrying := false
+
 		for {
-			if c.currentClient() == nil {
+			if retrying {
```
qiujian16, line 59: maybe have a clientReady() bool func, and check here?

pkg/cloudevents/generic/baseclient.go
```diff
@@ -152,18 +154,19 @@ func (c *baseClient) subscribe(ctx context.Context, receive receiveFn) {
 	// start a go routine to handle cloudevents subscription
 	go func() {
 		receiverCtx, receiverCancel := context.WithCancel(context.TODO())
-		cloudEventsClient := c.currentClient()
+		startReceiving := true
 
 		for {
-			if cloudEventsClient != nil {
+			if startReceiving {
```
qiujian16, line 160: why not use clientReady here also

pkg/cloudevents/generic/baseclient.go
skeeey: I think it may be not necessary, because the `retrying` flag always set after the client closed

pkg/cloudevents/generic/baseclient.go
skeeey: we always start the subscribe after the client connected, this is controlled by `restartReceiverSignal` channel in the base client internal, so I did not use the clientReady here

but for send part, it will be called by outside, so when we send an event, we need check the client if ready or not

pkg/cloudevents/generic/baseclient.go
skeeey: using client ready instead of `retrying` flag

pkg/cloudevents/generic/baseclient.go
```diff
@@ -123,14 +121,14 @@ func (c *baseClient) publish(ctx context.Context, evt cloudevents.Event) error {
 		return err
 	}
 
-	cloudEventsClient := c.currentClient()
-
-	if cloudEventsClient == nil {
+	c.RLock()
+	defer c.RUnlock()
+	if !c.clientReady {
```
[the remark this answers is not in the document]
skeeey: missed this part...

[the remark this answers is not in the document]
skeeey: done

PR #88, opened 2024-11-22
:bug: determine if ce informer store inited using informer hassynced

skeeey:
## Summary
Using informer hassynced  to determine if the store inited 
Fix the issue: 
When a source client send the events before the agent client informer is not synced, the event will not be resynced the events in the agent informer


qiujian16 (commented):
/approve
/lgtm

PR #89, opened 2025-02-11
:seedling: introduce the status hash extension

skeeey:
## Summary

Introduce the status hash extension in the status event, the source/broker may choose to save this hash to avoid recalculate the status hash


morvencao (commented):
lgtm

PR #91, opened 2025-02-13
:seedling: log the whole event with trace level

skeeey:
## Summary

log the whole event with trace level to avoid large output in debug level


morvencao (commented):
lgtm

qiujian16 (commented):
/approve
/lgtm

PR #92, opened 2025-02-14
🐛 fix inconsistent source label for cloudevent metrics

morvencao:
## Summary

the `source` label for cloudevents metrics is inconsistent between the server and agent:
on the agent, the `cloudevents_sent_total` metric has:
```
cloudevents_sent_total{source="maestro",cluster="cluster1",....}
```
on the server, the `cloudevents_received_total` metric has:
```
cloudevents_received_total{source="cluster1-work-agent",cluster="cluster1",....}
```
the agent’s source label should use the agent ID instead of "maestro" to match the server’s label.

## Related issue(s)

ref: https://issues.redhat.com/browse/ACM-17901


skeeey:
can we save the event source and event orignal source in the metrics, do you think it's valuable? @morvencao

morvencao:
> can we save the event source and event original source in the metrics together, do you think it's valuable? @morvencao

valuable, but only to cloudevents_sent_total from agent.  On server side, the original_source will be empty.

skeeey:
LGTM

PR #93, opened 2025-02-18
✨ Define grpc server interface

qiujian16:
## Related issue(s)

Fixes #

### DIFF
--- a/pkg/cloudevents/server/store.go
+++ b/pkg/cloudevents/server/store.go
@@ -1,0 +1,25 @@
+package server
+
+import (
+	"context"
+	cloudevents "github.com/cloudevents/sdk-go/v2"
+	cetypes "open-cluster-management.io/sdk-go/pkg/cloudevents/generic/types"
+)
+
+// Service is the interface that the Agent Event Server uses to get cloudevent from the backend storage,
+// sends to the related agent, and handle the statusUpdate event sent from the agent.
+
+// TODO need a method to check if an event has been processed already.
+type Service interface {
+	// Get the cloudEvent based on resourceID from the service
+	Get(ctx context.Context, resourceID string) (*cloudevents.Event, error)
+
+	// List the cloudEvent from the service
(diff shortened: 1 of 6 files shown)


pkg/cloudevents/server/mq/server.go
```diff
@@ -0,0 +1,122 @@
...
+
+	evt, err := service.Get(ctx, resourceID)
+	// if the resource is not found, it indicates the resource has been processed.
+	if errors.IsNotFound(err) {
+		return nil
+	}
+	if err != nil {
+		return err
+	}
+
```
skeeey, line 84: need to set cloudevent type to define the event type, like create/update/delete

```
eventType := types.CloudEventsType{
		CloudEventsDataType: t,
		SubResource:         "spec",
		Action:              types.EventAction("create_request"),
	}

evt.SetType(eventType.String())
```

pkg/cloudevents/server/store.go
```diff
@@ -0,0 +1,25 @@
+package server
+
+import (
+	"context"
+	cloudevents "github.com/cloudevents/sdk-go/v2"
+	cetypes "open-cluster-management.io/sdk-go/pkg/cloudevents/generic/types"
+)
+
+// Service is the interface that the GRPC broker uses to get cloudevent from the backend storage
```
morvencao, line 9: replace "the gRPC broker" with "Agent Event Server" since we now have an MQ implementation.

pkg/cloudevents/server/store.go
```diff
@@ -0,0 +1,25 @@
...
+// sends to the related agent, and handle the statusUpdate event sent from the agent.
+
+// TODO need a method to check if an event has been processed already.
+type Service interface {
```
morvencao, line 13: should we specify that the service implementation needs a queue for asynchronous event handling and error handling, including retries?

qiujian16: do we need to? Seems to me it depends...

morvencao: ok

PR #94, opened 2025-02-25
✨ enable certificate auto reload for grpc client.

morvencao:
## Summary

This PR enables the client certificate auto reload for the grpc cloudevents client, so that client doesn't need to restart when client certificate is rotated.

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/cloudevents/generic/options/grpc/options.go
+++ b/pkg/cloudevents/generic/options/grpc/options.go
@@ -3,7 +3,6 @@ package grpc
 import (
 	"context"
 	"crypto/tls"
-	"crypto/x509"
 	"fmt"
 	"os"
 	"time"
@@ -16,7 +15,6 @@ import (
 	"google.golang.org/grpc/credentials/oauth"
 	"google.golang.org/grpc/keepalive"
 	"gopkg.in/yaml.v2"
-	"k8s.io/klog/v2"
 
 	"open-cluster-management.io/sdk-go/pkg/cloudevents/generic/options"
 	"open-cluster-management.io/sdk-go/pkg/cloudevents/generic/options/cert"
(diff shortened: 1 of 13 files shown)


test/integration/cloudevents/broker/grpc.go
```diff
@@ -205,3 +219,46 @@ func (bkr *GRPCBroker) decode(evt *cloudevents.Event) (*store.Resource, error) {
...
+
+func ensureValidCertificateStream(srv any, stream grpc.ServerStream, info *grpc.StreamServerInfo, handler grpc.StreamHandler) error {
+	if err := validateClientCertificate(stream.Context()); err != nil {
+		return err
+	}
+
+	return handler(srv, stream)
+}
+
+func validateClientCertificate(ctx context.Context) error {
```
qiujian16, line 239: what if user use token only for authn, will this be ignored?

morvencao: yes.
client certificate authn and token authn are mutually exclusive by design.

morvencao: both client certificate authn and token authn require server-side TLS.

qiujian16: the func name is validateClientCertificate, so it is actual validating serving ca?

morvencao: no, it's validating client certificate.

pkg/cloudevents/generic/options/grpc/options.go
```diff
@@ -16,19 +16,23 @@ import (
...
+// GRPCDialer is a gRPC dialer that connects to a gRPC server
+// with the given URL, TLS configuration and keepalive options.
+type GRPCDialer struct {
 	URL              string
-	CAFile           string
-	ClientCertFile   string
-	ClientKeyFile    string
-	TokenFile        string
 	KeepAliveOptions KeepAliveOptions
+	TLSConfig        *tls.Config
```
skeeey, line 33: is is possible to provide an uniform util function to build this? then the mqtt and grcp can use it to build tls part?

pkg/cloudevents/generic/options/grpc/options.go
```diff
@@ -250,3 +263,29 @@ func (o *GRPCOptions) GetCloudEventsProtocol(ctx context.Context, errorHandler f
 	opts = append(opts, clientOpts...)
 	return protocol.NewProtocol(conn, opts...)
 }
+
+// rootCAs returns a cert pool to verify the TLS connection.
+// If the caFile is not provided, the default system certificate pool will be returned
+// If the caFile is provided, the provided CA will be appended to the system certificate pool
+func rootCAs(caFile string) (*x509.CertPool, error) {
```
skeeey, line 270: may put this in cert package as common function?

morvencao: done.

qiujian16:
/approve

also have we retreived user/group from the cert and set in ctx, so we can do authz later?

morvencao:
> also have we retreived user/group from the cert and set in ctx, so we can do authz later?
for the test grpc broker, we haven't add authz. but, yes, we can retreive user/group from cert and use it for authz.

PR #95, opened 2025-03-04
✨ support grpc subscription with data type.

morvencao:
## Summary

This PR adds support subscribing to cloudevents with specific datatype(s) via gRPC option.

## Related issue(s)

Fixes #


pkg/cloudevents/generic/options/grpc/protocol/option.go
```diff
@@ -11,6 +11,7 @@ type Option func(*Protocol) error
 type SubscribeOption struct {
 	Source      string
 	ClusterName string
+	DataTypes   []string // data type list for the client, eg. ["io.open-cluster-management.works.v1alpha1.manifests"]
```
qiujian16, line 14: I think we should only subscribe a single data type at one time, and I would consider here as a struct.

skeeey: This part is for grpc client, I think for one grpc client, it may subscribe multiple datatype resource with one connection, if we only subscribe a single data type at one time that means we will have multiple connections, does it make senses for grpc scenario?

morvencao: updated to single data type.

PR #96, opened 2025-03-05
🌱 remove manifest codec.

morvencao:
## Summary

This PR removes the manifest codec and its corresponding payload for single manifests. We now use the manifestbundle codec instead.

## Related issue(s)

Fixes #


skeeey:
LGTM

PR #97, opened 2025-03-07
🌱 single connection for grpc options.

morvencao:
## Summary

This PR ensures one single connection for each grpc/mqtt option.

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/cloudevents/generic/options/grpc/options.go
+++ b/pkg/cloudevents/generic/options/grpc/options.go
@@ -43,8 +43,10 @@ type KeepAliveOptions struct {
 
 // Dial connects to the gRPC server and returns a gRPC client connection.
 func (d *GRPCDialer) Dial() (*grpc.ClientConn, error) {
-	// Return the cached connection if it exists.
-	if d.conn != nil {
+	// Return the cached connection if it exists and is ready.
+	// Should not return a nil or unready connection, otherwise the caller (cloudevents client)
+	// will not use the connection for sending and receiving events in reconnect scenarios.
+	if d.conn != nil && d.conn.GetState() == connectivity.Ready {
 		return d.conn, nil
 	}
 	// Prepare gRPC dial options.
--- a/pkg/cloudevents/generic/options/grpc/options.go
+++ b/pkg/cloudevents/generic/options/grpc/options.go
@@ -43,6 +43,10 @@ type KeepAliveOptions struct {
 
 // Dial connects to the gRPC server and returns a gRPC client connection.
 func (d *GRPCDialer) Dial() (*grpc.ClientConn, error) {
+	// Return the cached connection if it exists.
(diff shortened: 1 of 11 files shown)


pkg/cloudevents/generic/options/grpc/options.go
```diff
@@ -43,6 +43,10 @@ type KeepAliveOptions struct {
 
 // Dial connects to the gRPC server and returns a gRPC client connection.
 func (d *GRPCDialer) Dial() (*grpc.ClientConn, error) {
+	// Return the cached connection if it exists.
```
qiujian16, line 46: should we add a lock here?

morvencao: `grpc.ClientConn` is thread safe from the doc: https://github.com/grpc/grpc-go/blob/master/Documentation/concurrency.md#clients

If we return a connection which is not ready, the base client will not get a new connection in reconnect cases, see:
https://github.com/open-cluster-management-io/sdk-go/blob/main/pkg/cloudevents/generic/baseclient.go#L63

qiujian16: please add some comments here.

qiujian16: what happens if the two goroutine are dialing at the same time?

morvencao: I misunderstood, the `GRPCDialer` needs a lock to protect the `conn` since multiple callers (cloudevents clients) may access it concurrently. Updated the code and comments.

pkg/cloudevents/generic/options/grpc/options.go
```diff
@@ -43,6 +45,15 @@ type KeepAliveOptions struct {
 
 // Dial connects to the gRPC server and returns a gRPC client connection.
 func (d *GRPCDialer) Dial() (*grpc.ClientConn, error) {
+	// Return the cached connection if it exists and is ready.
+	// Should not return a nil or unready connection, otherwise the caller (cloudevents client)
+	// will not use the connection for sending and receiving events in reconnect scenarios.
+	d.mu.RLock()
```
skeeey, line 51: I think we only lock once  at the beginning to avoid build the connection multiple times among the goroutines

morvencao: done.

skeeey:
LGTM

qiujian16:
could add an integration test that starts multiple thread and dial at the same time?

morvencao:
added an integration test::
1. two work agents start concurrently with the same gRPC options to ensure both can subscribe to the resource spec.
2. then, one agent is closed, causing its underlying gRPC connection to close.
3. the second work agent should automatically reconnect by recreating the gRPC connection.

PR #105, opened 2025-03-27
:sparkles: add cel library and common functions

haoqing0110:
## Related issue(s)

https://github.com/open-cluster-management-io/ocm/issues/843
https://github.com/open-cluster-management-io/ocm/issues/922


pkg/cel/common/values.go
```diff
@@ -0,0 +1,34 @@
...
+func ConvertObjectToUnstructured(obj interface{}) (*unstructured.Unstructured, error) {
+	if obj == nil || reflect.ValueOf(obj).IsNil() {
+		return &unstructured.Unstructured{Object: nil}, nil
+	}
+	ret, err := runtime.DefaultUnstructuredConverter.ToUnstructured(obj)
+	if err != nil {
+		return nil, err
+	}
+	return &unstructured.Unstructured{Object: ret}, nil
+}
```
bhperry, line 34: I think it's worthwhile to check if obj is already `*unstructured.Unstructured` here, which at least for the result of applying manifests will always be the case. `runtime.DefaultUnstructuredConverter.ToUnstructured` only checks if the obj implements the `runtime.Unstructured` interface, which `unstructured.Unstructured` does not.

I have a similar helper [in my WIP pr](https://github.com/open-cluster-management-io/ocm/pull/910/files#diff-1eb19addc11cd8fecd6bf310d649d8799c6676d69366230d21916e497f547f77R553-R564), will switch to this

haoqing0110: @bhperry thanks for this info. I add some UT for this func, could you help to check if case ["already unstructured"](https://github.com/open-cluster-management-io/sdk-go/pull/105/files#diff-02ce46e0efdc8761be944985bda6f7e682c91c70ac9c812221f7d7565470ef06R46) covers the case you mentioned? it seems the `unstructured.Unstructured` implements interface `runtime.Unstructured` and can be handled by `runtime.DefaultUnstructuredConverter.ToUnstructured`.

bhperry: Oh, my apologies. For some reason when I checked earlier I thought it didn't implement runtime.Unstructured, but you are right it does. Looks good then, thanks!

haoqing0110:
@bhperry yes! I meet the similar issue in OCM, and I downgraded the version of github.com/google/cel-go to v0.22.0 now. Plz try it again.

bhperry:
Nice, got the build working

PR #109, opened 2025-04-10
✨ grpc server options and authenticator

qiujian16:
## Related issue(s)

Fixes #


pkg/cloudevents/server/grpc/options/server.go
```diff
@@ -0,0 +1,214 @@
...
+	}))
+	grpcServerOptions = append(grpcServerOptions, grpc.KeepaliveParams(keepalive.ServerParameters{
+		MaxConnectionAge: s.options.MaxConnectionAge,
+		Time:             s.options.ServerPingInterval,
+		Timeout:          s.options.ServerPingTimeout,
+	}))
+
+	logger := klog.FromContext(ctx)
+	if s.options.TLSCertFile == "" && s.options.TLSKeyFile == "" {
+		cert, key, err := certutil.GenerateSelfSignedCertKey("localhost", []net.IP{net.ParseIP("127.0.0.1")}, nil)
```
skeeey, line 73: we may not need to do this? just return an error

PR #112, opened 2025-04-17
:bug: using cloudevent parseint to parse resource version

skeeey:
## Related issue(s)

Fixes #


qiujian16:
what is the diff from parseInt?

skeeey:
> what is the diff from parseInt?

parseInt can only convert string to int, this can accept any numeric value in int32 range, or canonical string

Using this because the in our code, the resourceverion extension may represent a string (cluster/lease/addon ... client) or a int (work client)

PR #113, opened 2025-04-21
:seedling: support StrategicMergePatchType

skeeey:
## Related issue(s)

Fixes #


qiujian16:
hrm, why do we need this?

skeeey:
for supporting the event patch

qiujian16:
could we add some comment on it is basically used in event? I am not quite sure if the stragty merge patch can be generally used.

skeeey:
add some explanations for stragty merge patch in the  patch function comments

PR #114, opened 2025-04-22
:sparkles: add CEL cost estimator

haoqing0110:
## Related issue(s)
https://github.com/open-cluster-management-io/ocm/issues/843
https://github.com/open-cluster-management-io/ocm/issues/955

### DIFF
--- a/pkg/cel/common/values.go
+++ b/pkg/cel/common/values.go
@@ -4,12 +4,15 @@ import (
 	"reflect"
 
 	"github.com/google/cel-go/cel"
+	"github.com/google/cel-go/common/types/ref"
 	"github.com/google/cel-go/ext"
+	"github.com/google/cel-go/interpreter"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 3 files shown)


pkg/cel/common/values.go
```diff
@@ -20,9 +23,25 @@ var BaseEnvOpts = []cel.EnvOption{
...
+	CostEstimator interpreter.ActualCostEstimator
+}
+
+// CallCost implements runtime cost estimation for CEL function calls
+func (b *BaseEnvCostEstimator) CallCost(function, overloadId string, args []ref.Val, result ref.Val) *uint64 {
+	k8sEstimator := &library.CostEstimator{}
+	if cost := k8sEstimator.CallCost(function, overloadId, args, result); cost != nil {
+		return cost
+	}
+	return b.CostEstimator.CallCost(function, overloadId, args, result)
```
zhujian7, line 41: I am not famillar with the CostEstimator, want to understand who is expected to assign value for the BaseEnvCostEstimator.CostEstimator? should we add the nil check for the `b.CostEstimator`?

haoqing0110: Yes! We should check that!

PR #116, opened 2025-05-13
:sparkles: Conditions CEL lib for easier status condition checks

bhperry:
## Summary

Adding ConditionsLib with a helper function to streamline status condition checks. Checking for the existence of conditions can be quite verbose with null checks at every level. The `hasConditions` function makes it easy to verify if a status map has any conditions before doing additional processing with `.exists(...)` or `.filter(...)`

## Related issue(s)
https://github.com/open-cluster-management-io/ocm/pull/910#discussion_r2075052856

### DIFF
--- a/pkg/cel/library/conditions.go
+++ b/pkg/cel/library/conditions.go
@@ -1,0 +1,71 @@
+package library
+
+import (
+	"github.com/google/cel-go/cel"
+	"github.com/google/cel-go/common/types"
+	"github.com/google/cel-go/common/types/ref"
+	"github.com/google/cel-go/common/types/traits"
+)
+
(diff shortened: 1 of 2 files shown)


pkg/cel/library/conditions.go
```diff
@@ -0,0 +1,71 @@
...
+func ConditionsLib() cel.EnvOption {
+	return cel.Lib(conditionsLib)
+}
+
+var conditionsLib = &conditionsLibType{}
+
+type conditionsLibType struct{}
+
+func (*conditionsLibType) LibraryName() string {
+	return "open-cluster-management.json"
```
haoqing0110, line 36: open-cluster-management.condition?

PR #118, opened 2025-05-21
:seedling: support certs data in config

skeeey:
## Summary

support specify certs data in config file

## Related issue(s)

Fixes #


pkg/cloudevents/generic/options/cert/config.go
```diff
@@ -0,0 +1,85 @@
...
+			return err
+		}
+		c.ClientKeyData = key
+	}
+
+	return nil
+}
+
+func (c *CertConfig) Validate() error {
+	if (len(c.ClientCertData) == 0 && len(c.ClientKeyData) != 0) ||
```
qiujian16, line 75: nit could use if (len(c.ClientCertData) == 0 ) == (len(c.ClientKeyData) == 0)) which is shorter.

skeeey: add nit check for each function

PR #140, opened 2025-09-10
 🐛 Add healthcheck for grpc server and client

qiujian16:
## Related issue(s)

Fixes #

## Summary by CodeRabbit

* **New Features**
  * Adds a gRPC heartbeat health endpoint and optional client-side health monitoring with configurable server heartbeat interval and client healthiness timeout.

* **Bug Fixes**
  * Health-watch now reports reconnect/timeout conditions non-blockingly to improve detection and recovery from unhealthy connections.

* **Style**
  * Minor internal naming and whitespace refinements.

* **Tests / Chores**
  * Added unit and integration tests, new config options/flags, and wiring to validate health-check behavior and defaults.

### DIFF
--- a/pkg/cloudevents/generic/options/grpc/protocol/protocol.go
+++ b/pkg/cloudevents/generic/options/grpc/protocol/protocol.go
@@ -33,7 +33,7 @@ type Protocol struct {
 
 	closeChan chan struct{}
 
-	// errorChan is to send an error messsage to restart the connection
+	// errorChan is to send an error message to restart the connection
 	reconnectErrorChan chan error
 	checkInterval      time.Duration
 }
@@ -184,6 +184,11 @@ func (p *Protocol) healthCheck(ctx context.Context) {
 	healthClient := healthpb.NewHealthClient(p.clientConn)
 	stream, err := healthClient.Watch(watchCtx, &healthpb.HealthCheckRequest{Service: ""})
 	if err != nil {
+		if s, ok := status.FromError(err); ok && s.Code() == codes.Unimplemented {
+			logger.Warnf("grpc-health Watch not implemented on server; skipping health check")
+			return
+		}
+		logger.Errorf("failed to watch health check, %v", err)
 		select {
 		case p.reconnectErrorChan <- err:
 		default:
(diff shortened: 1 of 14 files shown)


pkg/cloudevents/generic/options/grpc/protocol/protocol.go
```diff
@@ -164,3 +174,38 @@ func (p *Protocol) Close(ctx context.Context) error {
...
+		done := make(chan struct{})
+
+		go func() {
+			resp, err := stream.Recv()
+			if err != nil {
+				p.reconnectErrorChan <- err
+				close(done)
+				return
+			}
+			logger.Infof("Received server health status %s", resp.Status)
```
skeeey, line 199: do we need to handle the status? or we don't case about the status?

qiujian16: we do not, the status currently will only be a single value

pkg/cloudevents/generic/options/grpc/protocol/protocol.go
```diff
@@ -164,3 +174,64 @@ func (p *Protocol) Close(ctx context.Context) error {
...
+			select {
+			case healthReceived <- struct{}{}:
+			default:
+			}
+		}
+	}()
+
+	ticker := time.NewTicker(p.checkInterval)
```
clyang82, line 211: will checkInterval be equal to ServerHealthinessTimeout?

pkg/cloudevents/generic/options/grpc/protocol/protocol.go
```diff
@@ -164,3 +174,64 @@ func (p *Protocol) Close(ctx context.Context) error {
...
+
+// healthCheck check status of the server
+func (p *Protocol) healthCheck(ctx context.Context) {
+	logger := cecontext.LoggerFrom(ctx)
+	watchCtx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	healthClient := healthpb.NewHealthClient(p.clientConn)
+	stream, err := healthClient.Watch(watchCtx, &healthpb.HealthCheckRequest{Service: ""})
+	if err != nil {
```
morvencao, line 186: should we handle the watch error here for server `unimplemented` error? for example, in an upgrade scenario where the client is upgraded but the server is not, the client should not write an error to the error channel.

PR #144, opened 2025-09-15
:seedling: supporting heartbeat in cloudevents subscribe stream

skeeey:
## Related issue(s)

Fixes #

## Summary by CodeRabbit

- New Features
  - Configurable server healthiness timeout for gRPC connections.
  - Standard heartbeat event type and periodic heartbeat emission to subscribers.
  - Exposed reconnect-error channel with non-blocking reporting and fallback logging.
  - Asynchronous, buffered event delivery to subscribers for smoother performance.

- Bug Fixes
  - Improved handling of stream timeouts and reconnection to avoid stalls.
  - Safer error propagation when error channels are unavailable.

- Tests
  - Added in-memory gRPC tests validating heartbeat and timeout behaviors.

### DIFF
--- a/pkg/cloudevents/generic/options/grpc/options.go
+++ b/pkg/cloudevents/generic/options/grpc/options.go
@@ -114,6 +114,8 @@ func (d *GRPCDialer) Close() error {
 // GRPCOptions holds the options that are used to build gRPC client.
 type GRPCOptions struct {
 	Dialer *GRPCDialer
+
+	ServerHealthinessTimeout *time.Duration
 }
 
 // GRPCConfig holds the information needed to build connect to gRPC server as a given user.
@@ -130,6 +132,10 @@ type GRPCConfig struct {
 
 	// keepalive options
 	KeepAliveConfig KeepAliveConfig `json:"keepAliveConfig,omitempty" yaml:"keepAliveConfig,omitempty"`
+
+	// serverHealthinessTimeout is the max duration that client will reconnect if no server healthiness status is
+	// received in this duration, if it is not set, client will not reconnect when health message is received
+	ServerHealthinessTimeout *time.Duration `json:"serverHealthinessTimeout,omitempty" yaml:"serverHealthinessTimeout,omitempty"`
 }
 
 // KeepAliveConfig holds the keepalive options for the gRPC client.
@@ -241,6 +247,7 @@ func BuildGRPCOptionsFromFlags(configPath string) (*GRPCOptions, error) {
 		}
 	}
 
+	options.ServerHealthinessTimeout = config.ServerHealthinessTimeout
 	return options, nil
 }
 
--- a/pkg/cloudevents/generic/options/grpc/protocol/option.go
+++ b/pkg/cloudevents/generic/options/grpc/protocol/option.go
@@ -2,6 +2,7 @@ package protocol
 
 import (
 	"fmt"
+	"time"
 )
 
 // Option is the function signature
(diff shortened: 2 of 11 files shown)


pkg/cloudevents/generic/options/grpc/options.go
```diff
@@ -130,6 +132,10 @@ type GRPCConfig struct {
 
 	// keepalive options
 	KeepAliveConfig KeepAliveConfig `json:"keepAliveConfig,omitempty" yaml:"keepAliveConfig,omitempty"`
+
+	// serverHealthinessTimeout is the max duration that client will reconnect if no server healthiness status is
+	// received in this duration
+	ServerHealthinessTimeout time.Duration `json:"serverHealthinessTimeout,omitempty" yaml:"serverHealthinessTimeout,omitempty"`
```
qiujian16, line 138: make it as pointer, when it is not set, we should not start the timeout part

pkg/cloudevents/generic/options/grpc/protocol/option.go
```diff
@@ -24,3 +25,17 @@ func WithSubscribeOption(subscribeOpt *SubscribeOption) Option {
 		return nil
 	}
 }
+
+func WithServerHealthinessTimeout(timeout time.Duration, errorChan chan error) Option {
```
qiujian16, line 29: same here, the input is pointer. And if timeout <= 0, set a default value.

pkg/cloudevents/generic/options/grpc/protocol/protocol.go
```diff
@@ -123,17 +132,53 @@ func (p *Protocol) OpenInbound(ctx context.Context) error {
 
 	go func() {
 		for {
-			msg, err := subClient.Recv()
-			if err != nil {
+			msgCh := make(chan *pbv1.CloudEvent, 1)
+			errCh := make(chan error, 1)
+
+			go func() {
```
qiujian16, line 138: make it a goroutine independently.

```yaml
go func() {
for {
  msg, err := subClient.Recv()
  if err != nil {
      errCh <- err
	  return
  }
  msgCh <- msg
}
}
```

pkg/cloudevents/generic/types/types.go
```diff
@@ -9,6 +9,8 @@ import (
 	"github.com/google/uuid"
 )
 
+const HeartbeatCloudEventsType = "io.open-cluster-management.cloudevents.heartbeat"
```
qiujian16, line 12: add some comment

pkg/cloudevents/generic/options/grpc/options.go
```diff
@@ -130,6 +132,10 @@ type GRPCConfig struct {
 
 	// keepalive options
 	KeepAliveConfig KeepAliveConfig `json:"keepAliveConfig,omitempty" yaml:"keepAliveConfig,omitempty"`
+
+	// serverHealthinessTimeout is the max duration that client will reconnect if no server healthiness status is
+	// received in this duration
```
qiujian16, line 137: add if it is not set, client will not reconnect when health message is received

pkg/cloudevents/generic/options/grpc/protocol/protocol.go
```diff
@@ -30,6 +32,13 @@ type Protocol struct {
 	openerMutex sync.Mutex
 
 	closeChan chan struct{}
+
+	// reconnectErrorChan is to send an error message to reconnect the connection
+	reconnectErrorChan chan error
+
+	// serverHealthinessTimeout is the max duration that client will reconnect if no server healthiness
+	// status is received in this duration, by default it is 20s
+	serverHealthinessTimeout time.Duration
```
qiujian16, line 41: add if duration is 0, this will be disabled.

pkg/cloudevents/server/grpc/broker.go
```diff
@@ -181,14 +188,16 @@ func (bkr *GRPCBroker) Subscribe(subReq *pbv1.SubscriptionRequest, subServer pbv
 		return nil
 	})
 
+	go health.Heartbeat(ctx, subServer, bkr.heartbeatCheckInterval)
```
clyang82, line 191: Do we really need to start heartbeat in broker since it is only used in ACM context? @qiujian16

clyang82: maybe we have another PR to make it as an option.

pkg/cloudevents/generic/options/grpc/protocol/protocol.go
```diff
@@ -164,3 +167,77 @@ func (p *Protocol) Close(ctx context.Context) error {
...
+		for {
+			select {
+			case msg := <-heartbeatCh:
+				logger.Debugf("heartbeat received %v", msg)
+			case <-ctx.Done():
+				return
+			}
+		}
+	}
+
```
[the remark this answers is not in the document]
skeeey: when the error was sent, the client will be close `close(p.closeChan)`, this will trigger to  call cancel()

PR #224, opened 2026-05-27
:sparkles: Include metadata in WorkApplier cache to detect external changes

### LINKED ISSUE #223: WorkApplier cache does not detect external label/annotation changes
## Summary

`WorkApplier.Apply` caches the desired work hash and the existing work's generation to skip redundant updates via `safeToSkipApply`. However, metadata-only changes (labels, annotations) to a ManifestWork do not bump the `generation` field — only `resourceVersion` changes. This means `safeToSkipApply` returns true even when an external actor has modified labels/annotations that the applier owns, and `ManifestWorkEqual` is never called.

## Steps to reproduce

1. Use `WorkApplier.Apply` to create a ManifestWork with labels (e.g. `app.kubernetes.io/managed-by: my-addon`)
2. Externally modify one of those labels (e.g. change the value to `someone-else`)
3. Call `WorkApplier.Apply` again with the same desired work

**Expected:** `ManifestWorkEqual` detects the label diff and patches the MW back to the desired state.

**Actual:** `safeToSkipApply` returns true (same desired hash, same generation), the comparison is skipped entirely, and the tampered label persists.

## Analysis

The cache in `workcache.go` stores:

```go
type cachedResource struct {
    resourceHash string  // hash of desired work
    generation   int64   // generation of existing work
}
```

`safeToSkipApply` matches when both the desired hash and generation are unchanged. Since label changes don't bump generation (only spec changes do), the cache never invalidates for metadata-only modifications.

This breaks the reconciliation contract — a controller using `WorkApplier` cannot restore its owned labels/annotations if an external actor tampers with them.

## Environment

- sdk-go v1.3.0
- Discovered while integrating `WorkApplier` into an OCM addon controller

mkolesnik:
The WorkApplier cache uses the desired work hash and the existing work's generation to skip redundant updates. However, metadata-only changes (labels, annotations, owner references) do not bump generation, so the cache misses external modifications to these fields and `ManifestWorkEqual` is never called.

This stores a hash of the desired work's metadata in the cache and compares it against the existing work's metadata on each apply, ensuring the cache invalidates when an external actor modifies owned metadata fields.

Fixes: https://github.com/open-cluster-management-io/sdk-go/issues/223

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced metadata drift detection to identify changes in labels, annotations, and owner references, ensuring patches are issued when external modifications occur.

* **Tests**
  * Added comprehensive test coverage for metadata drift detection across multiple change scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/open-cluster-management-io/sdk-go/pull/224?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/pkg/apis/work/v1/applier/workcache.go
+++ b/pkg/apis/work/v1/applier/workcache.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"sync"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/klog/v2"
 
 	workapiv1 "open-cluster-management.io/api/work/v1"
@@ -20,6 +21,7 @@ type workKey struct {
 type cachedResource struct {
 	resourceHash string
 	generation   int64
+	metadataHash string
 }
 
 type workCache struct {
@@ -48,6 +50,7 @@ func (w *workCache) updateCache(required, existing *workapiv1.ManifestWork) {
 	value := cachedResource{
 		resourceHash: hashOfResourceStruct(required),
 		generation:   existing.Generation,
+		metadataHash: hashOfMetadata(required),
 	}
 
 	w.cache[key] = value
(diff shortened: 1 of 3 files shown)


pkg/apis/work/v1/applier/workcache.go
```diff
@@ -48,6 +50,7 @@ func (w *workCache) updateCache(required, existing *workapiv1.ManifestWork) {
 	value := cachedResource{
 		resourceHash: hashOfResourceStruct(required),
 		generation:   existing.Generation,
+		metadataHash: hashOfMetadata(required),
```
qiujian16, line 53: I think resourceHash has included metadata section already. Are you trying to detect if label/annotation of the exisiting mw is changed? If so, I would consider an annotatil/label merge when apply rather than replace.

mkolesnik: Thanks for the quick review!

You're right that `resourceHash` hashes the desired work, so it catches when we change our labels.
The problem is it doesn't catch when someone else changes labels on the existing work (see #223 which I reported).
When that happens, our `desired` hash stays the same and generation doesn't bump, so the cache skips and the external change is never corrected.

Re merge vs replace, that's a good point and worth discussing separately.
This PR just makes the cache consistent with what the applier already does when the cache misses.

mikeshng (commented):
@coderabbitai review

mikeshng (approved):
/lgtm

/assign @tesshuflower 

PTAL as well. Thanks!

PR #230, opened 2026-06-29
✨ Add ExtraDialOpts field to GRPCDialer

kuudori:
## Summary

Add `ExtraDialOpts []grpc.DialOption` field to `GRPCDialer`, allowing consumers to pass additional gRPC dial options that `GRPCDialer` does not configure directly (e.g. `grpc.WithConnectParams`, `grpc.WithAuthority`).

Options are appended last in `Dial()` — both TLS and insecure code paths — so caller-supplied options take precedence on conflict. Zero value (`nil`) preserves existing behavior.

**Use case:** configuring `grpc.WithConnectParams` to cap connection-level reconnect backoff (grpc-go defaults to `MaxDelay=120s`, which causes multi-minute recovery delays after server restarts).

## Related issue(s)

N/A — additive feature, no breaking changes.

## Summary by CodeRabbit

* **New Features**
  * Added support for passing custom gRPC dial settings when connecting.
  * Improved flexibility for gRPC connections by allowing additional connection options to be supplied.
* **Tests**
  * Added coverage for connecting with and without custom gRPC dial settings.

### DIFF
--- a/pkg/cloudevents/generic/options/grpc/options.go
+++ b/pkg/cloudevents/generic/options/grpc/options.go
@@ -32,6 +32,7 @@ type GRPCDialer struct {
 	KeepAliveOptions KeepAliveOptions
 	TLSConfig        *tls.Config
 	Token            string
+	ExtraDialOpts    []grpc.DialOption
 	mu               sync.Mutex       // Mutex to protect the connection.
 	conn             *grpc.ClientConn // Cached gRPC client connection.
 }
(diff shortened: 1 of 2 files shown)


tesshuflower:
@qiujian16  This looks pretty self-contained to me, but I don't seem to be able to approve - would you mind taking a look?

qiujian16:
this is weird. I think your id is in the owner list

PR #233, opened 2026-08-12
Fix agent watcher store write race that erases DeletionTimestamp

### LINKED ISSUE #232: Agent watcher store race: in-flight `Patch` can erase `DeletionTimestamp` and resurrect deleted ManifestWorks on the spoke
**Describe the bug**

There's a race between `ManifestWorkAgentClient.Patch` and the agent watcher store's `HandleReceivedResource`. When a delete event arrives while a status patch is in flight, the patch's final `watcherStore.Update` can write back a pre-delete copy of the work, erasing the `DeletionTimestamp` that `HandleReceivedResource` just set. The work controller's next requeue then finds a work with finalizers and no deletion timestamp, so it recreates the ManifestWork and AppliedManifestWork on the spoke, and from there it reapplies every 4 to 6 minutes indefinitely. We observed a zombie work being reapplied for 16+ hours until a spec resync happened to correct the store.

This looks like the root cause of the behaviour reported in open-cluster-management-io/ocm#1404.

**The race window**

`Patch` in `pkg/cloudevents/clients/work/agent/client/manifestwork.go` does roughly:

```
Get                 (~line 188)  reads the work at local RV N
Publish             (~line 230)  status event, network round trip
Get                 (~line 240)  second read
versionCompare      (~line 250)  patched RV vs store RV
watcherStore.Update (~line 253)  full object replacement in the cache
```

`HandleReceivedResource` is subscribed directly on the store (`options/generic.go`), so it doesn't share the client mutex that `Patch` holds. The double `versionCompare` actually catches most interleavings, because the agent work store bumps a local per-name resource version on every write. If the delete write lands before the second `Get`, the compare sees RV N+1 against patched RV N and returns a conflict, which is fine. The losing window is between the second `Get` and the `Update`:

1. `Patch` does its second `Get`, reads RV N, no deletion timestamp
2. `versionCompare` passes, N == N, only `newRV < lastRV` is rejected
3. `HandleReceivedResource` writes the deleting work at RV N+1
4. `Patch` calls `watcherStore.Update(newWork)`, the cache does a full object replacement, and the deletion timestamp is gone at RV N+2

The comment above the second `Get` already acknowledges the lack of read-modify-write guarantees in the store, this is a concrete reproduction of it biting.

The downstream effect is what makes it nasty. The finalize controller has usually already cleaned the spoke resources by this point using its own DeepCopy, so the spoke looks clean at first. But the finalizer removal never happens, `ManifestWorkFinalizeController` re-syncs, sees no deletion timestamp, and returns early. The work is now stuck as finalizer present plus no deletion timestamp, which passes the gate in the manifestwork controller sync, and `applyAppliedManifestWork` recreates everything on the next requeue.

**Observed timeline**

From a Maestro (openshift-online/maestro) agent using the gRPC driver, single work, one incident:

- `08:12:03.864` `create_request` received, spoke work applied
- `08:12:04` four status `update_request`s published in quick succession
- `08:12:06.886` `delete_request` received, `deletiontimestamp=08:12:06.726528Z`
- `08:12:07.090` spoke ManifestWork deleted ("is terminating"), so the store held the deletion timestamp at this point
- `08:12:14.662` status `update_request` published whose metadata has no `deletionTimestamp`, the store copy has lost it
- `08:17:33` requeue fires, spoke ManifestWork recreated with a new UID, plus a new AppliedManifestWork
- every 4 to 6 minutes after that, requeue and reapply, for roughly 16.5 hours
- `00:44:38` next day, a spec `resync_response` carrying the `deletiontimestamp` extension corrects the store and the work finally deletes

Happy to attach the full agent log.

**To Reproduce**

Delete a work within a few seconds of creating it, while status updates are still being published. The window itself is tiny but there are typically several status patches in flight right after create, we hit it 4 times in a week of e2e CI runs.

**Expected behaviour**

Once the delete event is processed, the store never again holds the work without a deletion timestamp, and the work converges to deleted.

**A second, related bug**

`AgentInformerWatcherStore.Delete` removes the versioner entry, so a deleted-then-recreated work restarts at RV 1. An in-flight stale `Patch` holding, say, RV 5 passes the `newRV < lastRV` check against RV 1 and can clobber a fresh work. Same root cause, the version check tolerates equality and versioner resets.

**Proposed fix**

We'd like to send a PR along these lines:

1. Add a mutex to the generic `AgentInformerWatcherStore[T]` guarding `Add`/`Update`/`Delete` and the whole `HandleReceivedResource` body.
2. Add an optional capability interface, something like `UpdateWithVersion(resource, expectedResourceVersion)`, implemented by the work agent store as a single critical section: get, compare resource version for equality, bump the versioner, write the store. Equality instead of `<` closes both bugs above.
3. `Patch` type-asserts for the capability and uses it for the final write, keeping the current Get/compare/Update as the fallback for other store implementations, so `ClientWatcherStore` doesn't change.

The CloudEvents publish stays outside the lock, so the lock is only ever held for an in-memory map operation, no contention on the network path. On conflict the caller already retries and re-derives the patch from the fresh store state, so nothing is lost.

**Environment ie: OCM version, Kubernetes version and provider:**

- `open-cluster-management.io/sdk-go` `v1.2.1-0.20260323031834-e885ccee3f1b`, and I've confirmed the relevant code is unchanged on current `main`
- Maestro server and agent over the gRPC driver, ARO HCP e2e environment

ciaranRoche:
Fixes #232

## What this does

Closes the write race between `ManifestWorkAgentClient.Patch` and the agent watcher store's `HandleReceivedResource` that erases `DeletionTimestamp` from the agent's in-memory store and resurrects deleted ManifestWorks on the spoke (full analysis and log timeline in #232, in the wild report in open-cluster-management-io/ocm#1404).

Three changes:

1. **Store-level mutex.** Both the generic `AgentInformerWatcherStore[T]` and the work agent store now serialize their writers, and `HandleReceivedResource` holds the lock for its whole read-modify-write sequence, so another writer can't be interleaved between reading the cached resource and writing it back.
2. **`ConditionalUpdater` capability.** A new optional interface in `clients/store` with one method, `UpdateWithVersion(ctx, resource, expectedResourceVersion)`. The work agent store implements it as a single critical section: get, compare the resource version for equality, bump the versioner, write. Equality (rather than the `<` check in `versionCompare`) also rejects stale writes after a delete/re-add cycle resets the per-name versioner, which was a second latent bug in the same code path.
3. **`Patch` uses the capability when available.** The final store write in `Patch` type-asserts for `ConditionalUpdater` and uses it, keeping the existing Get/versionCompare/Update as the fallback for stores that don't implement it, so `ClientWatcherStore` itself is unchanged and other implementations keep working as before.

The CloudEvents publish stays outside the lock, so the lock is only ever held for an in-memory map operation, there's no contention on the network path. On conflict the caller retries and re-derives the patch from the fresh store state (which now includes the deletion timestamp), so nothing is lost.

## Tests

- `TestManifestWorkAgentClient_ConcurrentStatusPatchAndDeleteEvent` reproduces the production failure. On current main it fails within a few iterations:

  ```
  --- FAIL: TestManifestWorkAgentClient_ConcurrentStatusPatchAndDeleteEvent (0.14s)
      manifestwork_race_test.go:225: iteration 34: the deletion timestamp was lost from the store
  ```

  With this change it passes reliably under `-race`.
- `TestManifestWorkAgentClient_Patch_DeleteEventDuringStatusPublish` deterministically interleaves a delete event while the status publish is in flight (via a publish hook) and asserts the stale patch gets a conflict and the deletion timestamp survives.
- `TestAgentInformerWatcherStore_UpdateWithVersion` covers the CAS semantics: match, stale, force with "0", empty version, missing work, the versioner reset case, and delete-event-wins.

`go test -race ./pkg/cloudevents/...` is green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01USPzrSgidrmV9YrxRRf5tQ

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented concurrent resource updates from overwriting one another.
  * Added atomic version checks to detect stale updates and return conflicts.
  * Preserved deletion information when status updates race with resource deletion.
  * Improved handling of missing resources and forced updates.
* **Reliability**
  * Strengthened event and cache consistency during simultaneous resource changes.
  * Added coverage for concurrent updates, deletion races, version conflicts, and resource recreation.

### DIFF
--- a/pkg/cloudevents/clients/store/informer.go
+++ b/pkg/cloudevents/clients/store/informer.go
@@ -3,6 +3,7 @@ package store
 import (
 	"context"
 	"reflect"
+	"sync"
 
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -23,6 +24,11 @@ import (
 type AgentInformerWatcherStore[T generic.ResourceObject] struct {
 	BaseClientWatchStore[T]
 	Watcher *Watcher
+
+	// mu synchronizes the store writers, making the read-modify-write sequence in
+	// HandleReceivedResource atomic with respect to the plain Add/Update/Delete writers
+	// (e.g. a resource client writing back a patched resource).
+	mu sync.Mutex
 }
 
 func NewAgentInformerWatcherStore[T generic.ResourceObject]() *AgentInformerWatcherStore[T] {
@@ -35,21 +41,44 @@ func NewAgentInformerWatcherStore[T generic.ResourceObject]() *AgentInformerWatc
 }
 
 func (s *AgentInformerWatcherStore[T]) Add(resource runtime.Object) error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	return s.add(resource)
+}
+
+func (s *AgentInformerWatcherStore[T]) Update(resource runtime.Object) error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	return s.update(resource)
+}
+
+func (s *AgentInformerWatcherStore[T]) Delete(resource runtime.Object) error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	return s.delete(resource)
+}
+
+func (s *AgentInformerWatcherStore[T]) add(resource runtime.Object) error {
 	s.Watcher.Receive(watch.Event{Type: watch.Added, Object: resource})
 	return s.Store.Add(resource)
 }
 
-func (s *AgentInformerWatcherStore[T]) Update(resource runtime.Object) error {
+func (s *AgentInformerWatcherStore[T]) update(resource runtime.Object) error {
 	s.Watcher.Receive(watch.Event{Type: watch.Modified, Object: resource})
 	return s.Store.Update(resource)
 }
 
-func (s *AgentInformerWatcherStore[T]) Delete(resource runtime.Object) error {
+func (s *AgentInformerWatcherStore[T]) delete(resource runtime.Object) error {
 	s.Watcher.Receive(watch.Event{Type: watch.Deleted, Object: resource})
 	return s.Store.Delete(resource)
 }
 
 func (s *AgentInformerWatcherStore[T]) HandleReceivedResource(ctx context.Context, resource T) error {
+	// Hold the store lock for the whole read-modify-write sequence, so another store writer
+	// cannot be interleaved between reading the cached resource and writing it back.
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
 	newRuntimeObj, err := utils.ToRuntimeObject(resource)
 	if err != nil {
 		return err
@@ -84,14 +113,14 @@ func (s *AgentInformerWatcherStore[T]) HandleReceivedResource(ctx context.Contex
 			if err != nil {
 				return err
 			}
-			return s.Update(cachedRuntimeObj)
+			return s.update(cachedRuntimeObj)
 		}
 
 		cachedRuntimeObj, err := utils.ToRuntimeObject(cachedMetaObj)
 		if err != nil {
 			return err
 		}
-		return s.Delete(cachedRuntimeObj)
+		return s.delete(cachedRuntimeObj)
 	}
 
 	_, exists, err := s.Get(ctx, newMetaObj.GetNamespace(), newMetaObj.GetName())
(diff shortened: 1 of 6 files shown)


pkg/cloudevents/clients/store/informer.go
```diff
@@ -35,21 +41,44 @@ func NewAgentInformerWatcherStore[T generic.ResourceObject]() *AgentInformerWatc
 }
 
 func (s *AgentInformerWatcherStore[T]) Add(resource runtime.Object) error {
+	s.mu.Lock()
```
qiujian16, line 44: I do not think work agent use this store. Do you need to update this generic store?

